PR #1, opened 2023-07-13
Post-repo-split work

jplatte:


src/events.rs
```diff
@@ -1,6 +1,6 @@
 //! Types related to events.
 
-use ruma::events::room::history_visibility::HistoryVisibility as RumaHistoryVisibility;
+use matrix_sdk_common::ruma::events::room::history_visibility::HistoryVisibility as RumaHistoryVisibility;
```
richvdh, line 3: bit surprised about this. Why is using `matrix_sdk_common`'s re-export better than using ruma directly?

richvdh (approved):
lgtm

src/events.rs
jplatte: Don't have to keep versions in sync that way. Especially relevant when we're using a git dependency of Ruma in the SDK (which I hope to change soon-ish, but not exactly high priority as far as EX is concerned).

richvdh: makes sense, thanks

PR #2, opened 2023-07-13
Rename to matrix-sdk-crypto-wasm

richvdh:
Since we're moving everything around anyway, let's take the opportunity to rename the project to matrix-sdk-crypto-wasm to better distinguish it from matrix-sdk-crypto-nodejs.


jplatte (approved):
💯

PR #4, opened 2023-07-13
Update README

richvdh:
Put some stuff about usage in the "Usage" section

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,6 @@ Encryption](https://en.wikipedia.org/wiki/End-to-end_encryption)) for
 
     or:
 
-
     ```
     yarn add @matrix-org/matrix-sdk-crypto-wasm
     ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,6 +13,41 @@ Encryption](https://en.wikipedia.org/wiki/End-to-end_encryption)) for
 
 ## Usage
 
+1. Install in your project:
+
+    ```
+    npm install --save @matrix-org/matrix-sdk-crypto-wasm
```
poljar, line 19: Can we stick to `yarn` or `npm`, but not have both?

poljar (approved):

README.md
richvdh: The problem is that you have to use the tool that the project you're building wants -- and `npm` is the lowest common denominator: people who use `yarn` can be expected to know how to translate an `npm` commandline into `yarn`, but `npm` users do not necessary know about `yarn`. Hence there's some sort of loose convention that "install this library" instructions use `npm`.

I can include yarn instructions as well.

PR #6, opened 2023-07-13
Release v1.0.0

richvdh:
Ensure we have a relesed version of matrix-sdk-crypto-wasm to point to.

This also updates the format of the release branch and tag names to be more conventional.


jplatte (approved):
Haven't really interacted with this crate or its release process before, but diff looks reasonable.

PR #8, opened 2023-07-13
GHA release workflow: fix permissions

richvdh:
Our new repo does not give write permissions by default. This is a good thing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,6 +48,7 @@ jobs:
 
       - name: Create the Github release
         uses: softprops/action-gh-release@v1
+        permissions:
+          contents: write
         with:
           draft: true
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -48,6 +48,7 @@ jobs:
 
       - name: Create the Github release
         uses: softprops/action-gh-release@v1
+        permissions:
+          contents: write
         with:
           draft: true
-          files: pkg/matrix-org-matrix-sdk-crypto-js-*.tgz
```
richvdh, line 53: there is not, nor has there ever been, such a file.

jplatte (approved):
I only saw in docs that per-job permissions are possible. If per-step works too, nice.

richvdh:
sigh I think you're right; it's per-workflow or per-job.

jplatte:
I don't think per-job is too bad ^^

PR #10, opened 2023-07-20
Expose WASM bindings for key backup

richvdh:
A re-do of https://github.com/matrix-org/matrix-rust-sdk/pull/2196, against the new repository, with a whole bunch of updates.

Closes: https://github.com/vector-im/crypto-internal/issues/100


AndrewFerr (changes requested):

tests/machine.test.js
```diff
@@ -911,4 +913,121 @@ describe(OlmMachine.name, () => {
...
+        });
+
+        test("accepts own signatures", async () => {
+            let m = await machine();
+            let _ = m.bootstrapCrossSigning(true);
+
+            let keyBackupKey = BackupDecryptionKey.createRandomKey();
+
+            let auth_data = {
+                public_key: keyBackupKey.megolmV1PublicKey.publicKeyBase64,
```
AndrewFerr, line 952: It's a bit confusing that the "decryption" key also contains the encryption key.

Case in point: the naming of `BackupDecryptionKey.createRandomKey()` suggests that it creates a random decryption key, when it actually creates a random public+private key pair.

Maybe `BackupDecryptionKey` (here & in [the PR from `matrix-rust-sdk`](https://github.com/matrix-org/matrix-rust-sdk/pull/2305)) should have been named just `BackupKey`?

Or this could be all moot if it's understood that generating a private key must also generate a public one to go along with it.

tests/backup.test.js
```diff
@@ -25,45 +25,22 @@ describe("BackupRecoveryKey", () => {
...
+                encryptedMegolm.session_data.ciphertext,
+            ),
         );
 
-        expect(decypted.algorithm).toStrictEqual(aMegolmKey.algorithm);
-        expect(decypted.sender_key).toStrictEqual(aMegolmKey.sender_key);
-        expect(decypted.session_key).toStrictEqual(aMegolmKey.session_key);
-    });
-
-    test("create export and import base58", () => {
```
AndrewFerr: It would be nice to keep support for base58 since some clients (namely Element Web) ask for the base58 format of the decryption key when restoring a backup.

tests/machine.test.js
```diff
@@ -911,4 +913,121 @@ describe(OlmMachine.name, () => {
...
+            let session_data = Object.values(sessions)[0].session_data;
+
+            // should decrypt with the created key
+            let decrypted = JSON.parse(
+                keyBackupKey.decryptV1(session_data.ephemeral, session_data.mac, session_data.ciphertext),
+            );
+            expect(decrypted.algorithm).toStrictEqual("m.megolm.v1.aes-sha2");
+
+            // simulate key backed up
+            m.markRequestAsSent(outgoing.id, outgoing.type, '{"etag":"1","count":3}');
```
AndrewFerr, line 1012:
```suggestion
            await m.markRequestAsSent(outgoing.id, outgoing.type, '{"etag":"1","count":3}');
```

tests/machine.test.js
richvdh: Yeah, it's not super-clear, I agree.

It's less that the decryption key *contains* the encryption key, and more that the encryption key can be *derived* from the decryption key, which is just the way that curve25519 keys work. `megolmV1PublicKey` here is actually a getter function which does some light maths to derive the public key.

so...

> Case in point: the naming of BackupDecryptionKey.createRandomKey() suggests that it creates a random decryption key, when it actually creates a random public+private key pair.

well no, it does actually create a random (private) decryption key. The public key is then derived later on.

tests/backup.test.js
richvdh: > some clients (namely Element Web) ask for the base58 format of the decryption key when restoring a backup.

I don't *think* they do, but I could well be missing something. Can you give me some more clues as to how to find this flow in the app?

richvdh:
@AndrewFerr thanks for the review! I think I've either accepted your suggestions or replied, so please could you take another look?

tests/machine.test.js
AndrewFerr: Alright, let's leave it as it is then.

tests/backup.test.js
AndrewFerr: The way to get this flow in Element Web is to use an account that skipped all identity/crypto/cross-signing setup (a brand new account is the easiest way to test that), then visit use Settings > Security & Privacy > Connect to Key Backup, which will have it ask for the base58 room backup decryption key instead of the 4S passphrase/key.

However, it's very easy to lose this flow, because once you've set up cross-signing / 4S, that button asks you for the 4S passphrase/key and there's seemingly no way to make it go back to asking for the backup decryption key.

AndrewFerr (approved):

poljar (approved):

PR #14, opened 2023-07-27
backups: expect a KeysBackupRequest

AndrewFerr:
Followup on https://github.com/matrix-org/matrix-rust-sdk/pull/2318

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -920,10 +920,10 @@ impl OlmMachine {
         future_to_promise(async move {
             match me.backup_machine().backup().await? {
                 Some((transaction_id, keys_backup_request)) => {
-                    Ok(Some(JsValue::from(requests::KeysBackupRequest::try_from((
+                    Ok(Some(requests::KeysBackupRequest::try_from((
                         transaction_id.to_string(),
                         &keys_backup_request,
-                    ))?)))
+                    ))?))
                 }
 
                 None => Ok(None),
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -918,16 +918,16 @@ impl OlmMachine {
         let me = self.inner.clone();
 
         future_to_promise(async move {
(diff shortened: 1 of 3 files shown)


src/machine.rs
```diff
@@ -918,16 +918,16 @@ impl OlmMachine {
...
 
         future_to_promise(async move {
-            let request = me.backup_machine().backup().await?;
-
-            let Some(request) = request else {
-                return Ok(None);
-            };
+            match me.backup_machine().backup().await? {
+                Some((transaction_id, keys_backup_request)) => {
+                    Ok(Some(JsValue::from(requests::KeysBackupRequest::try_from((
```
richvdh, line 923: not sure we need the explicit `JsValue::from` here? I think `future_to_promise` will work with anything which implements `Into<JsValue>` (as `KeybackupRequest does`) ?

src/machine.rs
```diff
@@ -918,16 +918,16 @@ impl OlmMachine {
         let me = self.inner.clone();
 
         future_to_promise(async move {
-            let request = me.backup_machine().backup().await?;
-
-            let Some(request) = request else {
-                return Ok(None);
-            };
+            match me.backup_machine().backup().await? {
```
richvdh, line 921: I preferred the `let else` over the `match` :stuck_out_tongue:.

YMMV, happy for you to change it.

src/machine.rs
AndrewFerr: Ah that's nice, I'll get rid of it then.

src/machine.rs
AndrewFerr: I went with the `match` to format this similarly to `OlmMachine.get_missing_sessions`, which is what gave me the idea of working with a `(request_id, request)` tuple in the first place.  Thanks for letting me keep it :slightly_smiling_face:

PR #19, opened 2023-08-04
Expose version field of KeysBackupRequest to bindings

BillCarsonFr:
It's needed in order to make the request in the js sdk (`OutgoingRequestProcessor`)


poljar (approved):
I think that this is good. At least the compiler agrees.

PR #22, opened 2023-08-23
feat: Add `future_to_promise_with_custom_error`

Hywan:
This patch implements `future_to_promise_with_custom_error`. It allows a `Future` to return a `JsValue` for the error, instead of returning a `JsError` directly. It provides more flexibility to return any kind of error types.


BillCarsonFr (approved):
LGTM, thx. Will push a PR next to use that to report typed decryption errors

PR #29, opened 2023-09-05
Expose new method `OlmMachine.getRoomEventEncryptionInfo`

richvdh:
Expose the new method added in https://github.com/matrix-org/matrix-rust-sdk/pull/2510

To do this, we need to give `EncryptionInfo` its own wrapper type; we can lift most of the functionality from `DecryptedRoomEvent`. And then we may as well reuse that wrapper in the `DecryptedRoomEvent`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # unreleased
 
+-   Add `OlmMachine.registerUserIdentityUpdatedCallback`.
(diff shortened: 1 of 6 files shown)


Hywan (approved):
Looks good, thanks!

richvdh:
I added a test, though it's a bit of a token effort if I'm totally honest.

Hywan:
Thanks!

PR #31, opened 2023-09-07
Fix resource leak in `registerUserIdentityUpdatedCallback`

richvdh:
Use the new `Store::identities_stream_raw` method introduced in https://github.com/matrix-org/matrix-rust-sdk/pull/2530 to avoid a resource leak.


poljar (approved):
Looks good.

PR #34, opened 2023-10-04
basic callback on secret stream

BillCarsonFr:
Adds support for secret stream in wasm.

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -1132,7 +1132,7 @@ impl OlmMachine {
     /// stored.
     ///
     /// Usually you would just register a callback with
-    /// [`register_receive_secret_callback`], but if the clients is shutdown
+    /// [`register_receive_secret_callback`], but if the client is shut down
     /// before handling them, this method can be used to retrieve them.
     ///
     /// The secrets are guaranteed to have been received over a 1-to-1 encrypted
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -1108,7 +1108,12 @@ impl OlmMachine {
             pin_mut!(stream);
             while let Some(secret) = stream.next().await {
                 send_secret_gossip_to_callback(callback_ref, &secret).await;
-                let _ = store.delete_secrets_from_inbox(&secret.secret_name).await;
+                match store.delete_secrets_from_inbox(&secret.secret_name).await {
+                    Ok(_) => (),
+                    Err(e) => {
+                        warn!("Error clearing secret inbox: {:?}", e);
+                    }
+                };
             }
         });
     }
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -1107,12 +1107,7 @@ impl OlmMachine {
             pin_mut!(stream);
             while let Some(secret) = stream.next().await {
                 send_secret_gossip_to_callback(&callback, &secret).await;
-                match store.delete_secrets_from_inbox(&secret.secret_name).await {
-                    Ok(_) => (),
-                    Err(e) => {
-                        warn!("Error clearing secret inbox: {:?}", e);
-                    }
-                }
+                let _ = store.delete_secrets_from_inbox(&secret.secret_name).await;
             }
         });
     }
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -1,6 +1,6 @@
 //! The crypto specific Olm objects.
 
-use std::{collections::BTreeMap, ops::Deref, time::Duration};
+use std::{collections::BTreeMap, ops::Deref, rc::Rc, sync::Arc, time::Duration};
 
 use futures_util::{pin_mut, StreamExt};
 use js_sys::{Array, Function, Map, Promise, Set};
@@ -1093,27 +1093,42 @@ impl OlmMachine {
             stream.for_each(move |item| send_user_identities_to_callback(callback_ref, item)).await;
         });
     }
-    /// Register a callback which will be called whenever a secret is received.
+
+    /// Register a callback which will be called whenever a secret
+    /// (`m.secret.send`) is received.
+    ///
+    /// To request a secret from other devices, a client sends an
+    /// `m.secret.request` device event with action set to request and name set
+    /// to the identifier of the secret. A device that wishes to share the
+    /// secret will reply with an m.secret.send event, encrypted using olm.
+    ///
+    /// The secrets are guaranteed to have been received over a 1-to-1 encrypted
+    /// to_device message from a verified own device.
+    ///
+    /// See https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_crypto/store/struct.Store.html#method.secrets_stream for more information.
     ///
     /// `callback` should be a function that takes 2 arguments the secret name
-    /// and value (base64) and returns a Promise.
+    /// and value and returns a Promise.
     #[wasm_bindgen(js_name = "registerReceiveSecretCallback")]
-    pub async fn register_secret_receive_callback(&self, callback: Function) {
+    pub async fn register_receive_secret_callback(&self, callback: Function) {
         let stream = self.inner.store().secrets_stream();
-        let store = self.inner.store().clone();
-        //fire up a promise chain which will call `cb` on each result from the stream
+        let store = Arc::new(self.inner.store().clone());
+        let store_weak = Arc::downgrade(&store);
+        // fire up a promise chain which will call `cb` on each result from the stream
         spawn_local(async move {
             let callback_ref = &callback;
             // Pin the stream to ensure it can be safely moved across threads
             pin_mut!(stream);
             while let Some(secret) = stream.next().await {
                 send_secret_gossip_to_callback(callback_ref, &secret).await;
-                match store.delete_secrets_from_inbox(&secret.secret_name).await {
-                    Ok(_) => (),
-                    Err(e) => {
-                        warn!("Error clearing secret inbox: {:?}", e);
-                    }
-                };
+                if let Some(store) = store_weak.upgrade() {
+                    match store.delete_secrets_from_inbox(&secret.secret_name).await {
+                        Ok(_) => (),
+                        Err(e) => {
+                            warn!("Error clearing secret inbox: {:?}", e);
+                        }
+                    };
+                }
             }
         });
     }
@@ -1143,8 +1158,8 @@ async fn send_room_key_info_to_callback(
     }
 }
 
-// helper for register_user_identity_updated_callback: passes the user ID into
-// the javascript function
+// helper for register_secret_receive_callback: passes the secret name and value
+// into the javascript function
 async fn send_user_identities_to_callback(
     callback: &Function,
     (identity_updates, _): (IdentityChanges, DeviceChanges),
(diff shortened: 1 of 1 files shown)


src/machine.rs
```diff
@@ -1093,6 +1093,30 @@ impl OlmMachine {
             stream.for_each(move |item| send_user_identities_to_callback(callback_ref, item)).await;
         });
     }
+    /// Register a callback which will be called whenever a secret is received.
+    ///
+    /// `callback` should be a function that takes 2 arguments the secret name
+    /// and value (base64) and returns a Promise.
+    #[wasm_bindgen(js_name = "registerReceiveSecretCallback")]
+    pub async fn register_secret_receive_callback(&self, callback: Function) {
```
richvdh, line 1101: please can you make these names match? Is it a "receive secret" callback or a "secret receive" callback?

richvdh (changes requested):
Any chance of a test?

src/machine.rs
```diff
@@ -1093,6 +1093,30 @@ impl OlmMachine {
             stream.for_each(move |item| send_user_identities_to_callback(callback_ref, item)).await;
         });
     }
+    /// Register a callback which will be called whenever a secret is received.
+    ///
+    /// `callback` should be a function that takes 2 arguments the secret name
+    /// and value (base64) and returns a Promise.
```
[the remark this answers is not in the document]
BillCarsonFr: Good point the spec don't talk about encoding, it's just a string

src/machine.rs
```diff
@@ -1093,6 +1093,30 @@ impl OlmMachine {
...
+    #[wasm_bindgen(js_name = "registerReceiveSecretCallback")]
+    pub async fn register_secret_receive_callback(&self, callback: Function) {
+        let stream = self.inner.store().secrets_stream();
+        let store = self.inner.store().clone();
+        //fire up a promise chain which will call `cb` on each result from the stream
+        spawn_local(async move {
+            let callback_ref = &callback;
```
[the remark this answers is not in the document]
BillCarsonFr: Yes me too, but the test fails if it is not used with "Did you forget to wait for something async in your test?" 
I don't understand why though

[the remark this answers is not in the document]
richvdh: which test?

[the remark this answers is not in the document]
BillCarsonFr: I don't know exactly what test, here is a failing run https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/actions/runs/6418397170/job/17426095599

But looks unrelated, it might just be flacky?

src/machine.rs
```diff
@@ -1094,6 +1094,44 @@ impl OlmMachine {
...
+    ///
+    /// See https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_crypto/store/struct.Store.html#method.secrets_stream for more information.
+    ///
+    /// `callback` should be a function that takes 2 arguments the secret name
+    /// and value and returns a Promise.
+    #[wasm_bindgen(js_name = "registerReceiveSecretCallback")]
+    pub async fn register_receive_secret_callback(&self, callback: Function) {
+        let stream = self.inner.store().secrets_stream();
+        let store = Arc::new(self.inner.store().clone());
+        let store_weak = Arc::downgrade(&store);
```
richvdh, line 1116:
```suggestion
        let store_weak = Arc::downgrade(self.inner.store());
```

I think?

src/machine.rs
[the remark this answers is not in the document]
richvdh: I would imagine it is a flaky test, yes

src/machine.rs
```diff
@@ -1094,6 +1094,44 @@ impl OlmMachine {
...
+        // fire up a promise chain which will call `cb` on each result from the stream
+        spawn_local(async move {
+            // Pin the stream to ensure it can be safely moved across threads
+            pin_mut!(stream);
+            while let Some(secret) = stream.next().await {
+                send_secret_gossip_to_callback(&callback, &secret).await;
+                if let Some(store) = store_weak.upgrade() {
+                    match store.delete_secrets_from_inbox(&secret.secret_name).await {
```
richvdh, line 1124: if the JS callback fails, we will still delete the secrets from the inbox. So we will lose the secret.

It is generally odd to me that registering a callback is enough that the secrets are automatically deleted.

BillCarsonFr: Changed, the js layer should now call delete secret.

src/machine.rs
```diff
@@ -1093,26 +1096,84 @@ impl OlmMachine {
...
+    /// secret will reply with an m.secret.send event, encrypted using olm.
+    ///
+    /// The secrets are guaranteed to have been received over a 1-to-1 encrypted
+    /// to_device message from a verified own device.
     ///
-    /// `callback` should be a function that takes 2 arguments the secret name
-    /// and value (base64) and returns a Promise.
+    /// See https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_crypto/store/struct.Store.html#method.secrets_stream for more information.
+    ///
+    /// `callback` should be a function that takes 2 arguments the secret name.
```
richvdh, line 1113: > `callback` should be a function that takes 2 arguments the secret name.

wat?

src/machine.rs
```diff
@@ -1093,26 +1096,84 @@ impl OlmMachine {
...
+    /// If the secret is valid and handled on the javascript side, the secret
+    /// inbox should be cleared by calling `delete_secrets_from_inbox`.
```
richvdh, line 1115: Suggest emphasising this a bit more:

```suggestion
    ///
    /// **Note**: if the secret is valid and handled on the javascript side, the secret
    /// inbox should be cleared by calling `delete_secrets_from_inbox`.
```

src/machine.rs
```diff
@@ -1093,26 +1096,84 @@ impl OlmMachine {
             stream.for_each(move |item| send_user_identities_to_callback(callback_ref, item)).await;
         });
     }
-    /// Register a callback which will be called whenever a secret is received.
+
+    /// Register a callback which will be called whenever a secret
+    /// (`m.secret.send`) is received.
+    ///
+    /// To request a secret from other devices, a client sends an
+    /// `m.secret.request` device event with action set to request and name set
```
richvdh, line 1104:
```suggestion
    /// `m.secret.request` device event with `action` set to `request` and `name` set
```

src/machine.rs
```diff
@@ -1093,26 +1096,84 @@ impl OlmMachine {
...
+    /// to the identifier of the secret. A device that wishes to share the
+    /// secret will reply with an m.secret.send event, encrypted using olm.
```
richvdh, line 1106:
```suggestion
    /// secret will reply with an `m.secret.send` event, encrypted using olm.
```

src/machine.rs
```diff
@@ -1093,26 +1096,84 @@ impl OlmMachine {
...
+    ///
+    /// The secrets are guaranteed to have been received over a 1-to-1 encrypted
+    /// to_device message from a verified own device.
```
richvdh, line 1109:
```suggestion
    /// The secrets are guaranteed to have been received over a 1-to-1 encrypted
    /// to_device message from a one of the user's own verified devices.
```

src/machine.rs
```diff
@@ -1093,26 +1096,84 @@ impl OlmMachine {
...
+                send_secret_gossip_to_callback(&callback, &secret).await;
             }
         });
     }
 
+    /// Get all the secrets with the given secret_name we have currently
+    /// stored.
+    ///
+    /// Usually you would just register a callback with
+    /// [`register_receive_secret_callback`], but if the clients is shutdown
```
richvdh, line 1134:
```suggestion
    /// [`register_receive_secret_callback`], but if the client is shut down
```

src/machine.rs
```diff
@@ -1093,26 +1096,84 @@ impl OlmMachine {
...
+    /// before handling them, this method can be used to retrieve them.
```
richvdh, line 1135:
```suggestion
    /// before handling them, this method can be used to retrieve them.
    /// This method should therefore be called at client startup to retrieve any
    /// secrets received during the previous session.
```

src/machine.rs
```diff
@@ -1093,26 +1096,84 @@ impl OlmMachine {
...
+    ///
+    /// The secrets are guaranteed to have been received over a 1-to-1 encrypted
+    /// to_device message from a verified own device.
```
richvdh, line 1138:
```suggestion
    /// The secrets are guaranteed to have been received over a 1-to-1 encrypted
    /// to_device message from one of the user's own verified devices.
```

src/machine.rs
```diff
@@ -1093,26 +1096,84 @@ impl OlmMachine {
             stream.for_each(move |item| send_user_identities_to_callback(callback_ref, item)).await;
         });
     }
-    /// Register a callback which will be called whenever a secret is received.
+
+    /// Register a callback which will be called whenever a secret
+    /// (`m.secret.send`) is received.
```
richvdh, line 1101: is it true that it is called for all `m.secret.send` events? I got the impression that some secrets (the cross-signing keys?) were handled internally? I might be wrong though

richvdh (commented):
a few nits & suggestions; generally seems good

BillCarsonFr:
> Any chance of a test?

There is no API to inject a secret to test the signaling. So no simple way to test it, but there is a full integrated test on the js-sdk side to test it

src/machine.rs
BillCarsonFr: No you are right for now it will only return is the backup decryption key. The cross signing keys are handled internally and custom secret are not yet supported

src/machine.rs
BillCarsonFr: udpated

src/machine.rs
```diff
@@ -1100,19 +1100,26 @@ impl OlmMachine {
...
-    /// to_device message from a verified own device.
+    /// to_device message from a one of the user's own verified devices.
     ///
     /// See https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_crypto/store/struct.Store.html#method.secrets_stream for more information.
     ///
-    /// `callback` should be a function that takes 2 arguments the secret name.
-    /// If the secret is valid and handled on the javascript side, the secret
-    /// inbox should be cleared by calling `delete_secrets_from_inbox`.
+    /// `callback` should be a function that takes 2 arguments: the secret name
+    /// (string) and value (string).
```
richvdh, line 1118: does it not matter what it returns?

src/machine.rs
```diff
@@ -1100,19 +1100,26 @@ impl OlmMachine {
     /// Register a callback which will be called whenever a secret
     /// (`m.secret.send`) is received.
     ///
+    /// The only secret this will currently broadcast is the
+    /// `m.megolm_backup.v1` (the cross signing secrets are handled internaly).
```
richvdh, line 1104:
```suggestion
    /// `m.megolm_backup.v1` (the cross signing secrets are handled internally).
```

src/machine.rs
```diff
@@ -1129,13 +1136,17 @@ impl OlmMachine {
 
     /// Get all the secrets with the given secret_name we have currently
     /// stored.
+    /// The only secret this will currently be returned is the
+    /// `m.megolm_backup.v1` secret.
```
richvdh, line 1140:
```suggestion
    /// The only secret this will currently return is the
    /// `m.megolm_backup.v1` secret.
```

richvdh (approved):
LGTM otherwise

richvdh:
(Looks like we have a flaky test)

BillCarsonFr:
> (Looks like we have a flaky test)

Yes, I see that we create OlmMachine but never close them?

PR #37, opened 2023-10-11
Update rust sdk

BillCarsonFr:
Update rust sdk to revision https://github.com/matrix-org/matrix-rust-sdk/commit/c2bb76029ae6d99c741727e0f87abcd734377016

Including the fixes:
[Remove dashmap crate usage from matrix-sdk-crypto](https://github.com/matrix-org/matrix-rust-sdk/pull/2669)]
[Bugfix for invalidated private cross signing keys not being persisted](https://github.com/matrix-org/matrix-rust-sdk/pull/2676)

Some breaking changes:
Following API changes on [Ruma](https://github.com/ruma/ruma/commit/984cbda9629f50c9ce2614270630971c42749895): 
  - Removed `localpart` and `server_name` method from `RoomId`.

 Rust SDK:
OlmMachine.sign returns a Result now

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,9 +40,9 @@ console_error_panic_hook = "0.1.7"
 futures-util = "0.3.27"
 http = "0.2.6"
 js-sys = "0.3.49"
-matrix-sdk-common = { git = "https://github.com/matrix-org/matrix-rust-sdk", rev = "880e918a5eaf66679f99fa902431d9d67f10cade", features = ["js"] }
-matrix-sdk-indexeddb = { git = "https://github.com/matrix-org/matrix-rust-sdk", rev = "880e918a5eaf66679f99fa902431d9d67f10cade" }
-matrix-sdk-qrcode = { git = "https://github.com/matrix-org/matrix-rust-sdk", rev = "880e918a5eaf66679f99fa902431d9d67f10cade", optional = true }
+matrix-sdk-common = { git = "https://github.com/matrix-org/matrix-rust-sdk", rev = "c2bb76029ae6d99c741727e0f87abcd734377016", features = ["js"] }
+matrix-sdk-indexeddb = { git = "https://github.com/matrix-org/matrix-rust-sdk", rev = "c2bb76029ae6d99c741727e0f87abcd734377016" }
+matrix-sdk-qrcode = { git = "https://github.com/matrix-org/matrix-rust-sdk", rev = "c2bb76029ae6d99c741727e0f87abcd734377016", optional = true }
 serde_json = "1.0.91"
 serde-wasm-bindgen = "0.5.0"
 tracing = { version = "0.1.36", default-features = false, features = ["std"] }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -40,19 +40,21 @@ console_error_panic_hook = "0.1.7"
...
+matrix-sdk-qrcode = { git = "https://github.com/matrix-org/matrix-rust-sdk", rev = "c2bb76029ae6d99c741727e0f87abcd734377016", optional = true }
 serde_json = "1.0.91"
 serde-wasm-bindgen = "0.5.0"
 tracing = { version = "0.1.36", default-features = false, features = ["std"] }
 tracing-subscriber = { version = "0.3.14", default-features = false, features = ["registry", "std"] }
 wasm-bindgen = "=0.2.84"
 wasm-bindgen-futures = "0.4.33"
 zeroize = "1.6.0"
+base64 = "0.13.0"
+rmp-serde = "1.1.2"
```
richvdh, line 54: do we particularly need these as direct dependencies now?

src/identifiers.rs
```diff
@@ -218,16 +218,10 @@ impl RoomId {
         Ok(Self::from(ruma::RoomId::parse(id)?))
     }
 
-    /// Returns the user's localpart.
-    #[wasm_bindgen(getter)]
-    pub fn localpart(&self) -> String {
-        self.inner.localpart().to_owned()
-    }
```
richvdh, line 225: removing this seems like a breaking change. Can we not maintain its behaviour somehow?

src/identifiers.rs
```diff
@@ -218,16 +218,10 @@ impl RoomId {
...
-
     /// Returns the server name of the room ID.
```
richvdh, line 221: this is no longer accurate.

Please please ensure that method behaviour is correctly documented. Failure to do so is one of the reasons matrix-js-sdk is such a nightmare to work on, and letting this library go down the same path would be terrible.

richvdh (changes requested):

src/identifiers.rs
BillCarsonFr: It's not used anywhere yet, so I was more keen to drop it and not having to maintain it?

I tried to use same reasoning that on the ruma change log

> Room IDs being splittable into localpart and servername does not have
much inherent value and there are proposals like MSC4051¹ that propose
changing the format. Relaxing the rules makes Ruma forwards-compatible
with those proposals. The server_name accessor is kept because it is
used by at least one downstream, but is updated to return an `Option`.

Cargo.toml
BillCarsonFr: I don't know what's the correct way to do it. If I don't add it the build is failing because it's not sure what version to use, but it's due to the Cargo.lock. 
I'll update the Cargo.lock instead.

src/identifiers.rs
richvdh: ok well, we better document it in the changelog as a breaking change.

BillCarsonFr: Good idea, I added a draft changelog.md for next version.

BillCarsonFr: As per our discussion I also removed RoomId.serverName

richvdh (approved):
Please can you update the PR summary now that we are removing server_name

> Changed server_name method on RoomId and RoomOrAliasId to return Option<&str>

LGTM otherwise

PR #40, opened 2023-10-17
Rewrite Tracing module

richvdh:
Rather than using a `Layer` which sends content to the JS console, construct an entire `Subscriber`, which allows us better control over what format is used, and includes data attached to spans.

### DIFF
--- a/src/tracing.rs
+++ b/src/tracing.rs
@@ -1,3 +1,11 @@
+use std::{
+    fmt,
+    sync::{Arc, Mutex, Once},
+};
+
+use ::tracing::Level;
+use matrix_sdk_common::js_tracing::{JsLoggingSubscriber, MakeJsLogWriter};
+use tracing_subscriber::{filter::LevelFilter, prelude::*, reload};
 use wasm_bindgen::prelude::*;
 
 /// Logger level.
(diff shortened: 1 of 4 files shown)


src/tracing.rs
```diff
@@ -1,3 +1,11 @@
+use std::{
+    fmt,
+    sync::{Arc, Mutex, Once},
+};
+
+use ::tracing::Level;
```
jplatte, line 6: The leading `::` should not be needed, since there is no local symbol called `tracing` in scope.

src/tracing.rs
```diff
@@ -31,257 +39,117 @@ pub enum LoggerLevel {
...
+}
 
-    struct Layer {
-        min_level: Level,
-        enabled: bool,
+#[wasm_bindgen]
+impl Tracing {
+    /// Check whether the `tracing` feature has been enabled.
+    #[wasm_bindgen(js_name = "isAvailable")]
+    pub fn is_available() -> bool {
```
jplatte, line 68: Kept for backwards-compatibility?

src/tracing.rs
```diff
@@ -31,257 +39,117 @@ pub enum LoggerLevel {
...
-        fn turn_on(&mut self) {
-            self.enabled = true;
-        }
-
-        fn turn_off(&mut self) {
-            self.enabled = false;
-        }
+    fn install_or_get_inner() -> Arc<Mutex<TracingInner>> {
+        static mut INSTALL: Option<Arc<Mutex<TracingInner>>> = None;
+        static INSTALLED: Once = Once::new();
```
jplatte, line 74: `OnceLock` is stable, can you not use that instead of a separate `Once` and `static mut`?

src/tracing.rs
richvdh: hrm, my IDE seemed to want to use the current file. Can try removing it.

src/tracing.rs
richvdh: yes, basically.

src/tracing.rs
richvdh: probably, this bit was cargo-culted from the previous impl! Will take a look.

jplatte (approved):

PR #42, opened 2023-10-18
Enable message-ids on to-device messages

richvdh:
Part of https://github.com/vector-im/element-web/issues/25722

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# UNRELEASED
+
+-   Attach message IDs to outgoing to-device messages, and log the IDs on
+    incoming messages.
+
 # matrix-sdk-crypto-wasm v2.0.0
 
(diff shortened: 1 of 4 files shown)


BillCarsonFr (approved):
LGTM.

I think we are not adding the msg ids "manually" in the js sdk right? So no changes need to be done in the js-sdk.

Also, can it be quickly tested in machine.test.js? when testing share room key.

```
            const content = JSON.parse(requests[0].body);
            expect(Object.keys(content.messages)).toEqual(["@example:localhost"]);

            const messageContent = content.messages["@example:localhost"]["AFGUOBTZWM"];
            expect(messageContent["org.matrix.msgid"]).toBeDefined();
```

richvdh:
> I think we are not adding the msg ids "manually" in the js sdk right? So no changes need to be done in the js-sdk.

It might be that we can improve the logging in the js-sdk.  I've opened https://github.com/matrix-org/matrix-js-sdk/pull/3810 already and am still working on the logging here.

But the message bodies themselves are passed straight through, so no need for changes otherwise.

> Also, can it be quickly tested in machine.test.js? 

yes good idea, thanks

PR #49, opened 2023-10-31
Update rust-sdk ver and `bootstrapCrossSigning` response

richvdh:
`bootstrapCrossSigning` now returns its own type, which can contain two or three nested requests. Create a new response type wrapper and return it.

Some other cleanups while we are here:

 * Get rid of `OutgoingRequests` wrapper, which was more confusing than helpful.

 * Rename `SigningKeysUploadRequest` to `UploadSigningKeysRequest` for consistency with matrix-rust-sdk.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,9 +39,9 @@ console_error_panic_hook = "0.1.7"
 futures-util = "0.3.27"
 http = "0.2.6"
 js-sys = "0.3.49"
-matrix-sdk-common = { git = "https://github.com/matrix-org/matrix-rust-sdk", rev = "7440ce0a0cfecf6c5e121e019470fde6ff38fefd", features = ["js"] }
-matrix-sdk-indexeddb = { git = "https://github.com/matrix-org/matrix-rust-sdk", rev = "7440ce0a0cfecf6c5e121e019470fde6ff38fefd" }
-matrix-sdk-qrcode = { git = "https://github.com/matrix-org/matrix-rust-sdk", rev = "7440ce0a0cfecf6c5e121e019470fde6ff38fefd", optional = true }
+matrix-sdk-common = { git = "https://github.com/matrix-org/matrix-rust-sdk", features = ["js"] }
+matrix-sdk-indexeddb = { git = "https://github.com/matrix-org/matrix-rust-sdk" }
+matrix-sdk-qrcode = { git = "https://github.com/matrix-org/matrix-rust-sdk", optional = true }
 serde_json = "1.0.91"
 serde-wasm-bindgen = "0.5.0"
 tracing = { version = "0.1.36", default-features = false, features = ["std"] }
@@ -55,6 +55,5 @@ vergen = { version = "8.0.0", features = ["build", "git", "gitcl"] }
 
 [dependencies.matrix-sdk-crypto]
 git = "https://github.com/matrix-org/matrix-rust-sdk"
-rev = "7440ce0a0cfecf6c5e121e019470fde6ff38fefd"
 default_features = false
 features = ["js", "backups_v1", "automatic-room-key-forwarding", "message-ids"]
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -39,9 +39,9 @@ console_error_panic_hook = "0.1.7"
 futures-util = "0.3.27"
 http = "0.2.6"
 js-sys = "0.3.49"
-matrix-sdk-common = { git = "https://github.com/matrix-org/matrix-rust-sdk", rev = "7440ce0a0cfecf6c5e121e019470fde6ff38fefd", features = ["js"] }
-matrix-sdk-indexeddb = { git = "https://github.com/matrix-org/matrix-rust-sdk", rev = "7440ce0a0cfecf6c5e121e019470fde6ff38fefd" }
-matrix-sdk-qrcode = { git = "https://github.com/matrix-org/matrix-rust-sdk", rev = "7440ce0a0cfecf6c5e121e019470fde6ff38fefd", optional = true }
```
bnjbvr, line 44: If you don't pin a revision, I'm worried this may try to use the latest one, and update the `Cargo.lock` for next users, so I'd recommend to keep on pinning a commit hash here (unless you've tested this and it doesn't change behavior even after new commits to upstream have landed).

src/requests.rs
```diff
@@ -29,7 +29,7 @@ use wasm_bindgen::prelude::*;
 /// Publishes end-to-end encryption keys for the device.
 ///
 /// [specification]: https://spec.matrix.org/unstable/client-server-api/#post_matrixclientv3keysupload
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
bnjbvr, line 32: here and below (x3 overall): is it useful to allow cloning these requests? Not sure how the homeserver will react if we replay them...

bnjbvr (approved):
LGTM!

bnjbvr:
(I just thought that an alternative design would be to create a large promise wrapping the three others: if first is present, run it and mark it as sync; then run second; then run third — and expose only that large promise to JS. It would simplify the end user API.)

Cargo.toml
richvdh: We discussed this in the sdk dev room (https://matrix.to/#/!qSsPTKDfMGYqhgiLPJ:flipdot.org/$Qd_cwgzB-POjqR7HicHup0jj-_n6GoxyTEAgmopBtkY?via=flipdot.org&via=matrix.org&via=element.io).

AIUI, `Cargo.lock` will only be updated if you do a `cargo update` or similar. I think that's fine.

src/requests.rs
richvdh: well, it's useful in that it's a requirement for them to be members of `CrossSigningBootstrapRequests`: the javascript-side getter for `upload_signing_keys_request` needs to be able to clone the `UploadSigningKeysRequest` object. We don't particularly plan to send the request twice, but making it clonable helps fit the rust ownership model to JS semantics.

That said: I now realise that `KeysUploadRequest` is not a member of `CrossSigningBootstrapRequests` (it was in an earlier attempt at this change). Will remove `Clone` for it.

richvdh:
```suggestion
#[derive(Debug)]
```

richvdh:
> (I just thought that an alternative design would be to create a large promise wrapping the three others: if first is present, run it and mark it as sync; then run second; then run third — and expose only that large promise to JS. It would simplify the end user API.)

I'm not really following this. We do need the details of the actual requests, because it's up to the JS side to actually make those requests over HTTP: rust-sdk-crypto-wasm itself does no networking.

bnjbvr:
> > (I just thought that an alternative design would be to create a large promise wrapping the three others: if first is present, run it and mark it as sync; then run second; then run third — and expose only that large promise to JS. It would simplify the end user API.)
> 
> I'm not really following this. We do need the details of the actual requests, because it's up to the JS side to actually make those requests over HTTP: rust-sdk-crypto-wasm itself does no networking.

Ah yes, i lacked that context, nevermind me.

PR #54, opened 2023-11-10
Expose new methods for importing room keys

richvdh:
Exposes the new import methods added in https://github.com/matrix-org/matrix-rust-sdk/pull/2759. Also adds a new return type which doesn't rely quite so much on JSON-encoded objects.

The old method is deprecated.

~~This PR currently based on #53.~~

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -823,6 +823,13 @@ impl OlmMachine {
     /// Mostly, a deprecated alias for `importExportedRoomKeys`, though the
     /// return type is different.
     ///
+    /// Returns a String containing a JSON-encoded object, holding three
+    /// properties:
+    ///  * `total_count` (the total number of keys found in the export data).
+    ///  * `imported_count` (the number of keys that were imported).
+    ///  * `keys` (the keys that were imported; a map from room id to a map of
+    ///    the sender key to a list of session ids).
+    ///
     /// @deprecated Use `importExportedRoomKeys` or `importBackedUpRoomKeys`.
     #[wasm_bindgen(js_name = "importRoomKeys")]
     pub fn import_room_keys(
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -3,16 +3,17 @@
 use std::{collections::BTreeMap, ops::Deref, time::Duration};
 
 use futures_util::{pin_mut, StreamExt};
-use js_sys::{Array, Function, Map, Promise, Set};
+use js_sys::{Array, Function, JsString, Map, Promise, Set};
 use matrix_sdk_common::ruma::{
     self, events::secret::request::SecretName, serde::Raw, DeviceKeyAlgorithm, OwnedTransactionId,
     UInt,
 };
 use matrix_sdk_crypto::{
     backups::MegolmV1BackupKey,
+    olm::BackedUpRoomKey,
     store::{DeviceChanges, IdentityChanges},
     types::RoomKeyBackupInfo,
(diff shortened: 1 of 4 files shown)


src/types.rs
```diff
@@ -227,3 +230,56 @@ impl SignatureVerification {
         self.inner.trusted()
     }
 }
+
+/// The result of a call to {@link OlmMachine.importExportedRoomKeys} or
+/// {@link OlmMachine.importBackedUpRoomKeys}.
+#[derive(Clone, Debug)]
+#[wasm_bindgen]
+pub struct RoomKeyImportResult {
+    /// The number of room keys that were imported.
```
bnjbvr, line 239: Maybe worth commenting how it might differ from `total_count`? (I've read that elsewhere in this PR, but since it's in a different file...)

src/types.rs
```diff
@@ -227,3 +230,56 @@ impl SignatureVerification {
...
+    /// Typescript type: `Map<string, Map<string, Set<string>>`.
+    pub fn keys(&self) -> Map {
+        let key_map = Map::new();
+
+        for (room_id, room_result) in self.keys.iter() {
+            let room_map = Map::new();
+            key_map.set(&JsString::from(room_id.to_string()), &room_map);
+
+            for (sender_key, sessions) in room_result.iter() {
+                let s: Array = sessions.iter().map(String::as_ref).map(JsString::from).collect();
```
bnjbvr, line 268: I think the two `map` calls could be fused into a single one? (less pretty, admittedly)

src/types.rs
```diff
@@ -227,3 +230,56 @@ impl SignatureVerification {
...
+                room_map.set(&JsString::from(sender_key.as_ref()), &Set::new(&s));
```
bnjbvr, line 269: Does the wasm binding for `Set::new` accept an iterator instead of an allocated collection? Mostly curious here (I think in JS `Set.new` will take any argument that implements the iteration protocol, e.g. a generator, and insert each value it yields into the set).

bnjbvr (approved):
Thanks! A long time I hadn't seen wasm binding code :star_struck:

src/types.rs
richvdh: I don't really know how it might differ, if I'm honest. I could hypothesise that it would exclude keys we already know about, but investigating this seems beyond the scope of this PR.

(This is just a copy of the documentation from https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk-crypto/src/lib.rs#L52.)

src/types.rs
richvdh: short answer: [not easily](https://github.com/rustwasm/wasm-bindgen/issues/1478), as far as I know.

src/types.rs
richvdh: yarp. Done.

src/machine.rs
```diff
@@ -819,34 +820,23 @@ impl OlmMachine {
 
     /// Import the given room keys into our store.
     ///
-    /// `exported_keys` is a list of previously exported keys that should be
-    /// imported into our store. If we already have a better version of a key,
-    /// the key will _not_ be imported.
+    /// Mostly, a deprecated alias for `importExportedRoomKeys`, though the
+    /// return type is different.
```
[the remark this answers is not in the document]
richvdh: Well, I didn't bother, because it wasn't documented before, and it's deprecated so there shouldn't be any *new* users who would need to know what it was. 

Have added some words to help a bit.

PR #55, opened 2023-11-10
Release v3.0.1

richvdh:


richvdh:
don't think this should be contentious; merging without review.

PR #58, opened 2023-11-25
Update dependencies

richvdh:


src/tracing.rs
```diff
@@ -78,13 +78,7 @@ impl Tracing {
...
         INSTALL
             .get_or_init(|| {
-                let format = tracing_subscriber::fmt::format().without_time().pretty();
-                let subscriber = tracing_subscriber::fmt()
-                    .with_max_level(Level::TRACE)
-                    .with_writer(MakeJsLogWriter::new())
-                    .with_ansi(false)
-                    .event_format(format)
-                    .finish();
+                let subscriber = make_tracing_subscriber(None);
```
richvdh, line 81: https://github.com/matrix-org/matrix-rust-sdk/pull/2856 changed the type of `JsLoggingSubscriber`, and we may as well just use `make_tracing_subscriber` rather than re-inventing it here.

src/js.rs
```diff
@@ -13,7 +13,7 @@ where
     let constructor_name = Object::get_prototype_of(value).constructor().name();
 
     if constructor_name == classname {
-        let pointer = Reflect::get(value, &JsValue::from_str("ptr"))
+        let pointer = Reflect::get(value, &JsValue::from_str("__wbg_ptr"))
```
richvdh, line 16: cf https://github.com/rustwasm/wasm-bindgen/pull/3408

jplatte: This seems rather brittle but I guess the wbg dependency is pinned anyways 🥲

jplatte (approved):

src/js.rs
richvdh: totally agree. 

We need to get rid of this function anyway, per https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/51

PR #62, opened 2023-12-01
New accessors `roomKeyRequestsEnabled` and `roomKeyForwardingEnabled`

richvdh:
Follow-up to https://github.com/matrix-org/matrix-rust-sdk/pull/2902. Required for https://github.com/vector-im/element-web/issues/26524.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,9 +1,2 @@
-[patch.'https://github.com/matrix-org/matrix-rust-sdk']
-matrix-sdk-base = { path = "../matrix-rust-sdk/crates/matrix-sdk-base" }
-matrix-sdk-common = { path = "../matrix-rust-sdk/crates/matrix-sdk-common" }
-matrix-sdk-crypto = { path = "../matrix-rust-sdk/crates/matrix-sdk-crypto" }
-matrix-sdk-indexeddb = { path = "../matrix-rust-sdk/crates/matrix-sdk-indexeddb" }
-matrix-sdk-qrcode = { path = "../matrix-rust-sdk/crates/matrix-sdk-qrcode" }
-
 [build]
 target = "wasm32-unknown-unknown"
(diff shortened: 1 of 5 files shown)


.cargo/config
```diff
@@ -1,2 +1,9 @@
+[patch.'https://github.com/matrix-org/matrix-rust-sdk']
+matrix-sdk-base = { path = "../matrix-rust-sdk/crates/matrix-sdk-base" }
+matrix-sdk-common = { path = "../matrix-rust-sdk/crates/matrix-sdk-common" }
+matrix-sdk-crypto = { path = "../matrix-rust-sdk/crates/matrix-sdk-crypto" }
+matrix-sdk-indexeddb = { path = "../matrix-rust-sdk/crates/matrix-sdk-indexeddb" }
+matrix-sdk-qrcode = { path = "../matrix-rust-sdk/crates/matrix-sdk-qrcode" }
```
bnjbvr, line 6: This might not build for everyone (including CI)?

src/machine.rs
```diff
@@ -172,6 +172,38 @@ impl OlmMachine {
...
+    /// have missed in the original share using `m.room_key_request`
+    /// events.
+    #[wasm_bindgen(getter, js_name = "roomKeyRequestsEnabled")]
+    pub fn are_room_key_requests_enabled(&self) -> bool {
+        self.inner.are_room_key_requests_enabled()
+    }
+
+    /// Enable or disable automatic transmission of room key requests.
+    #[wasm_bindgen(setter, js_name = "roomKeyRequestsEnabled")]
+    pub fn toggle_room_key_requests(&self, enabled: bool) {
```
bnjbvr, line 187: (here and below for the other function) uber-nit: admittedly I'm being a bit pedantic here, but to me `toggle` carries the meaning of setting the value to `!value`. Would it make sense to replace the prefix with `set_` or `enable_` instead?

tests/machine.test.js
```diff
@@ -201,6 +201,20 @@ describe(OlmMachine.name, () => {
         expect(await machine().displayName).toBeUndefined();
     });
 
+    test("can toggle room key requests", async () => {
+        const m = await machine();
+        expect(m.roomKeyRequestsEnabled).toBeTruthy();
```
bnjbvr, line 206: Curious, why do we not use true and false in these tests? We're not dealing with null, undefined et al., so a triple-equal comparison would return what we expect.

.cargo/config
richvdh: oh sorry, that wasn't meant to be here.

PR #63, opened 2023-12-01
Expose requestMissingSecrets API in bindings

BillCarsonFr:
Quick bindings for rust `query_missing_secrets_from_other_sessions`

### DIFF
--- a/tests/machine.test.js
+++ b/tests/machine.test.js
@@ -1134,4 +1134,177 @@ describe(OlmMachine.name, () => {
             expect(progressListener).toHaveBeenCalledWith(0, 1);
         });
     });
+
+    describe("Request missing secrets", () => {
+        async function getHypotheticalKeyQuery(new_user, new_device, bootstrapCrossSigning) {
+            let initialMachine = await OlmMachine.initialize(new_user, new_device);
+            const userId = initialMachine.userId.toString();
+            const deviceId = initialMachine.deviceId.toString();
+
+            let deviceKeys;
+            let outgoingRequests = await initialMachine.outgoingRequests();
+            for (const request of outgoingRequests) {
+                if (request instanceof KeysUploadRequest) {
+                    deviceKeys = JSON.parse(request.body);
+                }
+            }
+            delete deviceKeys.one_time_keys;
+            delete deviceKeys.fallback_keys;
+
+            let xsigning;
+
+            if (bootstrapCrossSigning) {
+                let bootstrapRequest = await initialMachine.bootstrapCrossSigning(true);
+
+                xsigning = JSON.parse(bootstrapRequest[0].body);
+                const newSignature = JSON.parse(bootstrapRequest[1].body);
+
+                console.log("Device keys", JSON.stringify(deviceKeys));
+                console.log("Xsigning", JSON.stringify(xsigning));
+                console.log("newSignature", JSON.stringify(newSignature));
+
+                const allSignatures = {
+                    [userId]: {
+                        ...deviceKeys.device_keys.signatures[userId],
+                        ...newSignature[userId][deviceId].signatures[userId],
+                    },
+                };
+
+                deviceKeys.device_keys.signatures = allSignatures;
+            }
+
+            return {
+                device_keys: {
+                    [userId]: {
+                        [deviceId]: deviceKeys,
+                    },
+                },
+                ...xsigning,
+            };
+        }
+
+        async function getSecondMachine(userId, deviceId, hypothetical_response) {
+            const secondMachine = await OlmMachine.initialize(userId, deviceId);
+
+            const toDeviceEvents = JSON.stringify([]);
+            const changedDevices = new DeviceLists();
+            const oneTimeKeyCounts = new Map();
+            const unusedFallbackKeys = new Set();
+
+            await secondMachine.receiveSyncChanges(
+                toDeviceEvents,
+                changedDevices,
+                oneTimeKeyCounts,
+                unusedFallbackKeys,
+            );
+            let outgoingRequests = await secondMachine.outgoingRequests();
+
+            const request = outgoingRequests[1];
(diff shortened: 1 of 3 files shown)


tests/machine.test.js
```diff
@@ -1134,4 +1134,177 @@ describe(OlmMachine.name, () => {
...
+            let outgoingRequests = await initialMachine.outgoingRequests();
+            for (const request of outgoingRequests) {
+                if (request instanceof KeysUploadRequest) {
+                    deviceKeys = JSON.parse(request.body);
+                }
+            }
+            delete deviceKeys.one_time_keys;
+            delete deviceKeys.fallback_keys;
+
+            let xsigning;
```
richvdh, line 1154: we don't really use this abbreviation anywhere else. Suggest calling it `crossSigning` and sucking up the extra 4 characters.

tests/machine.test.js
```diff
@@ -1134,4 +1134,177 @@ describe(OlmMachine.name, () => {
             expect(progressListener).toHaveBeenCalledWith(0, 1);
         });
     });
+
+    describe("Request missing secrets", () => {
+        async function getHypotheticalKeyQuery(new_user, new_device, bootstrapCrossSigning) {
```
richvdh, line 1139: I think it would be good to document what this function does, what its parameters mean, what it returns, etc.

Also doesn't it return a *response*, rather than a *query*? So it probably needs renaming.

tests/machine.test.js
```diff
@@ -1134,4 +1134,177 @@ describe(OlmMachine.name, () => {
...
+
+            if (bootstrapCrossSigning) {
+                let bootstrapRequest = await initialMachine.bootstrapCrossSigning(true);
+
+                xsigning = JSON.parse(bootstrapRequest[0].body);
+                const newSignature = JSON.parse(bootstrapRequest[1].body);
+
+                console.log("Device keys", JSON.stringify(deviceKeys));
+                console.log("Xsigning", JSON.stringify(xsigning));
+                console.log("newSignature", JSON.stringify(newSignature));
```
richvdh, line 1164: doesn't it log the right thing without `JSON.stringify`?

Also, this looks like it might be rather noisy?

tests/machine.test.js
```diff
@@ -1134,4 +1134,177 @@ describe(OlmMachine.name, () => {
...
+                device_keys: {
+                    [userId]: {
+                        [deviceId]: deviceKeys,
+                    },
+                },
+                ...xsigning,
+            };
+        }
+
+        async function getSecondMachine(userId, deviceId, hypothetical_response) {
```
richvdh, line 1186: again, please document this function

tests/machine.test.js
```diff
@@ -1134,4 +1134,177 @@ describe(OlmMachine.name, () => {
...
+                    console.log("ToDeviceRequest", request.body);
+                }
+            }
+
+            expect(mskRequested).toStrictEqual(true);
+            expect(sskRequested).toStrictEqual(true);
+            expect(uskRequested).toStrictEqual(true);
+        });
+
+        test("Should not request if there are request already in flight", async () => {
```
richvdh, line 1253:
```suggestion
        test("Should not request if there are requests already in flight", async () => {
```

richvdh (changes requested):
looks sensible but the tests and docs could do with a cleanup

tests/machine.test.js
BillCarsonFr: yes forgot to remove them

tests/machine.test.js
BillCarsonFr: Updated name and doc. Regarding the naming convention, I am not sure what should be used in this project there is a mix of snake and camel case.
Regarding `bootstrapCrossSigning` I removed it

tests/machine.test.js
```diff
@@ -1134,4 +1134,177 @@ describe(OlmMachine.name, () => {
...
+                            uskRequested = true;
+                        }
+                    }
+                    console.log("ToDeviceRequest", request.body);
+                }
+            }
+
+            expect(mskRequested).toStrictEqual(false);
+            expect(sskRequested).toStrictEqual(false);
+            expect(uskRequested).toStrictEqual(false);
```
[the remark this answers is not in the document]
BillCarsonFr: Sure done it. Wanted to make the test more resistent to future changes where outgoingRequests would contain new things. But could be done later I guess

tests/machine.test.js
richvdh: any use of snake case in JS code is a bug, imho

tests/machine.test.js
```diff
@@ -1150,8 +1150,22 @@ describe(OlmMachine.name, () => {
...
 
     describe("Request missing secrets", () => {
-        async function getHypotheticalKeyQuery(new_user, new_device, bootstrapCrossSigning) {
-            let initialMachine = await OlmMachine.initialize(new_user, new_device);
+        /**
+         *
+         * Creates an hypothetical response to a key query request for an account with a pre-existing device and identity.
+         *
+         * To be used in test when you want to create a setup where there is an existing device
+         * on the account with cross-signing setup.
```
richvdh, line 1158: Nit:

```suggestion
         * on the account with cross-signing set up.
```

["setup" is a noun. "set up" is the verb]

tests/machine.test.js
```diff
@@ -1225,10 +1243,10 @@ describe(OlmMachine.name, () => {
             return secondMachine;
         }
 
-        test("Should request cross signing request if missing", async () => {
+        test("Should request cross signing-keys if missing", async () => {
```
richvdh, line 1246: sorry, this was my bad:

```suggestion
        test("Should request cross-signing keys if missing", async () => {
```

tests/machine.test.js
BillCarsonFr: resolved as a separate PR

PR #71, opened 2023-12-12
handle verification cancel codes as strings

uhoreg:


richvdh:
Context: currently, cancel codes are exposed as integers, which isn't very useful. Per https://github.com/matrix-org/matrix-js-sdk/pull/3947, we think the raw strings would make more sense.

src/verification.rs
```diff
@@ -621,8 +619,8 @@ impl CancelInfo {
 
     /// Get the `CancelCode` that cancelled this verification.
```
richvdh, line 620: no longer a `CancelCode`. It would be nice to document what it actually *is*, with an example.

richvdh (commented):
Looks sensible to me.

Would be nice to improve documentation of the methods - what does a cancel code actually look like?

Would also be nice to add a test, though that may be a palaver that I won't insist on.

We could also do with an entry in the changelog.

Hywan (commented):
So, the idea is to rely on the `From<&str> for CancelCode` implementation of Ruma instead of exposing our own “proxy” `CancelCode` enum. Why not.

However, if the `code` (as a `&str`) is “invalid”, it will be transformed into `CancelCode::_Custom`. Is it something we want? It will basically hide all potential errors.

uhoreg:
> However, if the `code` (as a `&str`) is “invalid”, it will be transformed into `CancelCode::_Custom`. Is it something we want? It will basically hide all potential errors.

I would expect that `cancelWithCode` would usually get called with a hard-coded string, rather than with the output of a function, so I don't think it will be an issue.  (Currently, `cancelWithCode` isn't called anywhere, so I don't think we need to worry too much about it).

src/verification.rs
uhoreg: done.  also added changelog and tests

Hywan (approved):

PR #72, opened 2023-12-14
Correctly encode Duration and optional fields

uhoreg:
fixes https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/56, https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/57, and (hopefully) https://github.com/element-hq/element-web/issues/26566

I ported https://github.com/matrix-org/matrix-rust-sdk/pull/817 over, and added another parameter to mark fields as optional if they're `Option<...>`, to avoid sticking `null` in responses.  It's kind of ugly, so suggestions welcome.

I'm not sure what's the best way to test this.  All the existing tests are in JS, and as far as I can tell, from JS the only way to get these structs is to get them from OlmMachine, which means coercing OlmMachine to spit out the right requests.  Otherwise, I could write a test in Rust that creates the ruma request struct and test that it transforms into the wasm request struct correctly, but that would mean adding another testing system.  That's not too bad since Rust has its own testing built-in, but I don't know if we care about keeping the tests in JS.


src/requests.rs
```diff
@@ -316,7 +318,7 @@ macro_rules! request {
     (
         $destination_request:ident from $source_request:ident
         $( extracts $( $field_name:ident : $field_type:tt ),+ $(,)? )?
-        $( $( and )? groups $( $grouped_field_name:ident ),+ $(,)? )?
+        $( $( and )? groups $( $grouped_field_name:ident $( { $transformation:expr } )? $( $optional:literal )? ),+ $(,)? )?
```
uhoreg, line 321: I'm adding `"optional"` to flag things as optional.  But the match will match any literal, which is suboptimal.  I'd prefer to match just `optional`, but I don't see any way of doing that.  If I use `$( optional )`, then that would match just `optional`, but then I don't have a way to expand that back out it the transcription below, because it isn't assigned to a metavariable.

uhoreg:
> Otherwise, I could write a test in Rust that creates the ruma request struct and test that it transforms into the wasm request struct correctly, but that would mean adding another testing system. That's not too bad since Rust has its own testing built-in, but I don't know if we care about keeping the tests in JS.

Just for fun, I decided to see what would happen if I ran `cargo test`, and I got

```
    Finished test [unoptimized + debuginfo] target(s) in 56.12s
     Running unittests src/lib.rs (target/wasm32-unknown-unknown/debug/deps/matrix_sdk_crypto_wasm-a46182b35ca3089b.wasm)
error: test failed, to rerun pass `--lib`

Caused by:
  could not execute process `/home/hubert/Projects/matrix/matrix-rust-sdk-crypto-wasm/target/wasm32-unknown-unknown/debug/deps/matrix_sdk_crypto_wasm-a46182b35ca3089b.wasm` (never executed)

Caused by:
  Exec format error (os error 8)
```

which suggests that doing Rust tests will not be as trivial as I thought it would be.

uhoreg:
I figured out a way to test the conversions, and it's only moderately ugly!

src/requests.rs
```diff
@@ -619,3 +640,45 @@ impl TryFrom<OriginalCrossSigningBootstrapRequests> for CrossSigningBootstrapReq
...
+            upload_keys::v3::Request as OriginalKeysUploadRequest,
+        },
+        device_id, user_id, DeviceKeyAlgorithm,
+    };
+    use matrix_sdk_crypto::requests::KeysQueryRequest as OriginalKeysQueryRequest;
+    use wasm_bindgen::prelude::wasm_bindgen;
+
+    use super::{KeysClaimRequest, KeysQueryRequest, KeysUploadRequest};
+
+    #[wasm_bindgen(js_name = "_test_make_keys_claim_request")]
```
uhoreg, line 661: I deliberately used ugly names here since they're only intended for testing

src/requests.rs
andybalaam: Could we check at runtime that the token is `optional`? It would be nice to have some kind of check.

andybalaam: Or I guess you could have two alternatives, one with `optional` and one without, and give the metavariables different names? I've not done many of these macros so I'm just guessing.

andybalaam (approved):
I found these macros very difficult to understand, but as far as I did understand, this looks sensible, and the tests appear to demonstrate it works :-)

src/requests.rs
uhoreg: @Hywan as original author of the `request!` macro, do you have any suggestions for this?

src/requests.rs
```diff
@@ -619,3 +640,45 @@ impl TryFrom<OriginalCrossSigningBootstrapRequests> for CrossSigningBootstrapReq
...
+            user_id!("@alice:localhost").to_owned(),
+            BTreeMap::from([(
+                device_id!("ABCDEFG").to_owned(),
+                DeviceKeyAlgorithm::SignedCurve25519,
+            )]),
+        )]));
+        KeysClaimRequest::try_from(("ID".to_string(), &request)).unwrap()
+    }
+
+    #[wasm_bindgen(js_name = "_test_make_keys_query_request")]
```
Hywan, line 673: It's too bad that it's going to be part of the public API :-/. Can't we use https://docs.rs/wasm-bindgen-test/ to address that?

src/requests.rs
Hywan: What about simply `optional`?

```rust
$( $( and )? groups $( $grouped_field_name:ident $( { $transformation:expr } )? $( optional )? ),+ $(,)? )?
```

Hywan (changes requested):

src/requests.rs
uhoreg: Yup, I tried doing `$( optional )?`, but I couldn't figure out how to handle the expansion so that it does something different depending on whether it's specified or not, because it isn't bound to a metavariable.

src/requests.rs
uhoreg: done

Cargo.toml
```diff
@@ -70,6 +70,9 @@ tracing-subscriber = { version = "0.3.14", default-features = false, features =
 wasm-bindgen = "=0.2.87"
 wasm-bindgen-futures = "0.4.33"
 zeroize = "1.6.0"
+# 0.3.37 is the latest version that works with wasm-bindgen 0.2.87.  When the
+# dependency on wasm-bindgen is relaxed, this can be relaxed as well
+wasm-bindgen-test = "=0.3.37"
```
Hywan, line 75: Please rebase on top of `main`, it uses `wasm-bindgen` 0.2.89 now.

src/requests.rs
Hywan: Let's keep things as they are now, I may update this later.

Hywan (changes requested):

Cargo.toml
uhoreg: The [`Cargo.toml`](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/blob/main/Cargo.toml#L70) still says 0.2.87

richvdh: that's because https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/82 still hasn't landed yet.

uhoreg: Merged `main` and bumped the `wasm-bindgen-test` version.

Hywan (approved):

PR #76, opened 2023-12-29
Create new `StoreHandle` class to save repeatedly opening stores

richvdh:
A prerequisite for supporting migration from libolm: opening the indexeddb takes a long time, so we don't want to do it for each step in the migration.

### DIFF
--- a/src/store.rs
+++ b/src/store.rs
@@ -1,11 +1,107 @@
 //! Store types.
 
+use std::sync::Arc;
+
+use matrix_sdk_crypto::store::{DynCryptoStore, IntoCryptoStore, MemoryStore};
 use wasm_bindgen::prelude::*;
 
 use crate::{
     encryption::EncryptionAlgorithm, identifiers::RoomId, impl_from_to_inner,
     vodozemac::Curve25519PublicKey,
 };
+
+/// A struct containing an open connection to a CryptoStore.
+///
+/// Opening the CryptoStore can take some time, due to the PBKDF calculation
+/// involved, so if multiple operations are being done on the same store, it is
+/// more efficient to open it once.
+#[wasm_bindgen]
+#[derive(Clone, Debug)]
+pub struct StoreHandle {
+    pub(crate) store: Arc<DynCryptoStore>,
+}
+
+#[wasm_bindgen]
+impl StoreHandle {
+    /// Open a crypto store.
+    ///
+    /// The created store will be based on IndexedDB if a `store_name` is
+    /// provided; otherwise it will be based on a memory store and once the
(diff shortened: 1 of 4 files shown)


tests/machine.test.ts
```diff
@@ -53,15 +54,28 @@ describe("Versions", () => {
...
-    test("can be instantiated with a store", async () => {
+    test("can be instantiated with a StoreHandle", async () => {
         let storeName = "hello";
         let storePassphrase = "world";
 
+        let storeHandle = await StoreHandle.open(storeName, storePassphrase);
+        expect(
+            await OlmMachine.init_from_store(new UserId("@foo:bar.org"), new DeviceId("baz"), storeHandle),
+        ).toBeInstanceOf(OlmMachine);
+        storeHandle.free();
```
BillCarsonFr, line 72: Is it needed to call it explicitly?

BillCarsonFr (approved):
LGTM

tests/machine.test.ts
bnjbvr: There are no concept of deterministic finalizers in JavaScript (the only finalizers that exist are non-deterministic due to the nature of GC).

richvdh: Yeah. In theory the finalizer should get run soon after `storeHandle` goes out of scope; in practice it seemed to take a while for the garbage collector to clean it up. Without this, the tests ran slowly.

src/store.rs
```diff
@@ -1,12 +1,108 @@
...
+    #[wasm_bindgen(js_name = "open")]
+    pub async fn open_for_js(
+        store_name: Option<String>,
+        store_passphrase: Option<String>,
+    ) -> Result<JsValue, JsValue> {
+        Ok(StoreHandle::open(store_name, store_passphrase)
+            .await
+            .map_err(|e| JsError::from(&*e))?
+            .into())
+    }
```
Hywan, line 51: Ideally, you may want to split the Rust API from the JS API like so:

```rust
// The Rust API
impl StoreHandle {
    pub(crate) async fn open_(…) { … } // Note the name: `open_`
    // …
}

// The JS API
#[wasm_bindgen]
impl StoreHandle {
    pub async fn open(…) { … }
    // …
}
```

richvdh: @Hywan I don't quite follow. Are you asking that:
 * I put the two functions into separate `impl` blocks, with `#[wasm_bindgen]` on only one of them? If so, why is this an improvement?
 * I rename the functions? Happy to do so, but it feels very much like a matter of taste?

richvdh: Assuming this wasn't a blocking comment, so merging.

PR #77, opened 2023-12-29
Support for importing stored data from libolm

richvdh:
Fixes https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/78

~~Based on https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/76~~


src/libolm_migration.rs
```diff
@@ -0,0 +1,436 @@
...
+        sender_key,
+        signing_key: iter::once((
+            DeviceKeyAlgorithm::Ed25519,
+            Ed25519PublicKey::from_base64(&libolm_session.sender_signing_key)?.into(),
+        ))
+        .collect(),
+        room_id: libolm_session
+            .room_id
+            .clone()
+            .ok_or(JsError::new("Room ID not specified for megolm session"))?
```
bnjbvr, line 406: `.context` here too

tests/libolm_migration.test.ts
```diff
@@ -0,0 +1,79 @@
...
+            "testPass",
+        );
+        expect(olmMachine.identityKeys.curve25519.toBase64()).toEqual("LKv0bKbc0EC4h0jknbemv3QalEkeYvuNeUXVRgVVTTU");
+        expect(olmMachine.identityKeys.ed25519.toBase64()).toEqual("qK70DEqIXq7T+UU3v/al47Ab4JkMEBLpNrTBMbS5rrw");
+
+        const backupKeys: BackupKeys = await olmMachine.getBackupKeys();
+        expect(backupKeys.backupVersion).toEqual("3");
+        expect(backupKeys.decryptionKey?.toBase64()).toEqual("/FLbqTHzH1ihmQl3740Dm2aWgOzBng8HjYdGuCpuMLU");
+
+        // TODO: figure out a way to test cross-signing key import
```
bnjbvr, line 77: Is this something you wanted to address before landing this?

bnjbvr (changes requested):
Sweet! Mostly rust-isms in my comments, it would be super nice if someone from the crypto team could double-check the crypto logic here.

src/libolm_migration.rs
```diff
@@ -0,0 +1,436 @@
...
+pub struct BaseMigrationData {
+    /// The user id of the account owner.
+    #[wasm_bindgen(js_name = "userId")]
+    pub user_id: Option<UserId>,
+
+    /// The device ID of the account owner.
+    #[wasm_bindgen(js_name = "deviceId")]
+    pub device_id: Option<DeviceId>,
+
+    /// The pickled version of the Olm Account.
```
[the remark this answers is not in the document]
richvdh: I don't think it's helpful to dive into the gory details of the format (it's an encrypted blob containing a binary structure which is nominally a libolm implementation detail), but I've updated the comment to hopefully better explain where it comes from.

src/libolm_migration.rs
```diff
@@ -0,0 +1,436 @@
...
+        Ok(JsValue::UNDEFINED)
+    }
+}
+
+async fn migrate_base_data_to_store(
+    data: &BaseMigrationData,
+    pickle_key: &[u8],
+    store: &DynCryptoStore,
+) -> anyhow::Result<()> {
+    let user_id = data.user_id.clone().ok_or(anyhow!("User ID not specified"))?.inner;
```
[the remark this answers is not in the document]
richvdh: Nice tip about the `Option` extension, thanks!

> Alternatively, since the user_id is mandatory in the BaseMigrationData, let's make it a non Optional field?

Well... that would mean we'd have to pass in a UserID in the constructor of `BaseMigrationData`, and once we go down that path, pretty soon we get to a situation where the constructor takes 150 parameters. The whole point of `BaseMigrationData` is to avoid functions that take lots of parameters and instead allow the javascript side to use named properties.

src/libolm_migration.rs
```diff
@@ -0,0 +1,436 @@
...
+    let account = vodozemac::olm::Account::from_libolm_pickle(&data.pickled_account, pickle_key)?;
+    let account =
+        matrix_sdk_crypto::olm::Account::from_pickle(matrix_sdk_crypto::olm::PickledAccount {
+            user_id: user_id.clone(),
+            device_id: data.device_id.clone().ok_or(anyhow!("Device ID not specified"))?.inner,
+            pickle: account.pickle(),
+            shared: true,
+            // assume we have 50 keys on the server, until we get a sync that says fewer.
+            uploaded_signed_key_count: 50,
```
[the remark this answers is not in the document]
richvdh: Well, first of all: it only matters if we have run out of one-time keys (OTKs) on the server; in which case, yes, until we receive the first `/sync` response, we won't upload any new keys. But it's worth noting that the fact we're starting our application doesn't make much difference here: most likely, we ran out of OTKs before the app even started. Waiting a few more minutes until we get the first `/sync` response won't make much difference here.

Secondly: the matrix protocol handles the "out of OTKs" situation by having a "fallback key", which is used as a last resort once OTKs have run out (and is replaced once the application starts and discovers the fallback key has been used). It has slightly inferior security properties to a proper OTK, but means that running out of OTKs isn't a disaster.

src/libolm_migration.rs
```diff
@@ -0,0 +1,436 @@
...
+
+/// A pickled version of a `Session`.
+///
+/// Holds all the information that needs to be stored in a database to restore
+/// an Olm Session. Can be imported into the rust store with {@link
+/// #migrateOlmSessions}.
+#[wasm_bindgen(getter_with_clone)]
+#[derive(Debug)]
+pub struct PickledSession {
+    /// The libolm pickle string holding the Olm Session.
```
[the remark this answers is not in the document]
richvdh: again, have updated the comment.

src/libolm_migration.rs
```diff
@@ -0,0 +1,436 @@
...
+            .collect::<Result<_>>()?;
+
+        import_olm_sessions_to_store(rust_sessions, store_handle.store.as_ref())
+            .await
+            .map_err(|e| JsError::from(&*e))?;
+        Ok(JsValue::UNDEFINED)
+    }
+}
+
+impl Default for PickledSession {
```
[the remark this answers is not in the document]
richvdh: I might be missing something, but `Date` doesn't implement `Default`, so `derive(Default)` doesn't work.

src/libolm_migration.rs
```diff
@@ -0,0 +1,436 @@
...
+    libolm_session: JsValue,
+    pickle_key: &[u8],
+) -> Result<matrix_sdk_crypto::olm::PickledSession> {
+    let libolm_session = downcast::<PickledSession>(&libolm_session, "PickledSession")?;
+    let session = vodozemac::olm::Session::from_libolm_pickle(&libolm_session.pickle, &pickle_key)?;
+
+    let creation_time = date_to_seconds_since_epoch(&libolm_session.creation_time)
+        .ok_or(JsError::new("session creation time out of range"))?;
+    let last_use_time = date_to_seconds_since_epoch(&libolm_session.last_use_time)
+        .ok_or(JsError::new("session last-use time out of range"))?;
```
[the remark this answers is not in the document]
richvdh: noted. updated.

src/libolm_migration.rs
```diff
@@ -0,0 +1,436 @@
...
+}
+
+/// A pickled version of an `InboundGroupSession`.
+///
+/// Holds all the information that needs to be stored in a database to restore
+/// an InboundGroupSession.
+#[wasm_bindgen(getter_with_clone)]
+#[derive(Debug, Default)]
+pub struct PickledInboundGroupSession {
+    /// The liblolm pickle string holding the InboundGroupSession.
```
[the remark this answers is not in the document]
richvdh: as before: updated the comment.

src/libolm_migration.rs
richvdh: this method returns `Result<_, JsError>`, rather than `std::Error` or whatever, so I don't think `.context` works here. Have switched to `.ok_or_else` though.

tests/libolm_migration.test.ts
richvdh: Not really, tbh. It looks *tricky*.

richvdh:
@bnjbvr: Thanks for the review! Please take another look.

src/libolm_migration.rs
```diff
@@ -0,0 +1,436 @@
...
+        .map(|k| BackupDecryptionKey::from_base64(k.as_str()))
+        .transpose()?;
+
+    let cross_signing = PrivateCrossSigningIdentity::empty(&user_id);
+    cross_signing
+        .import_secrets_unchecked(
+            data.cross_signing_master_key.as_deref(),
+            data.cross_signing_self_signing_key.as_deref(),
+            data.cross_signing_user_signing_key.as_deref(),
+        )
```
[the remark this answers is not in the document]
bnjbvr: (Yes I meant that, modulo order of words :sweat_smile:) Thanks for the explanations!

src/libolm_migration.rs
```diff
@@ -0,0 +1,440 @@
...
+    ///   store the vodozemac data.
+    #[wasm_bindgen(js_name = "migrateOlmSessions")]
+    pub async fn migrate_olm_sessions(
+        sessions: Array,
+        pickle_key: Uint8Array,
+        store_handle: &StoreHandle,
+    ) -> Result<JsValue, JsError> {
+        let rust_sessions = sessions
+            .into_iter()
+            .map(|s| libolm_pickled_session_to_rust_pickled_session(s, &(pickle_key.to_vec())))
```
bnjbvr, line 234: Just seeing this: it might be possible that `pickle_key.to_vec()` could be hoisted above the `let rust_sessions = ` statement, to avoid creating the pickle key vector for every single session.

src/libolm_migration.rs
```diff
@@ -0,0 +1,440 @@
...
+    ///   store the vodozemac data.
+    #[wasm_bindgen(js_name = "migrateMegolmSessions")]
+    pub async fn migrate_megolm_sessions(
+        sessions: Array,
+        pickle_key: Uint8Array,
+        store_handle: &StoreHandle,
+    ) -> Result<JsValue, JsError> {
+        let rust_sessions = sessions
+            .into_iter()
+            .map(|s| libolm_pickled_megolm_session_to_rust_pickled_session(s, &(pickle_key.to_vec())))
```
bnjbvr, line 374: Ditto here

src/libolm_migration.rs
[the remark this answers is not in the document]
bnjbvr: I think there's a middle ground where the two required parameters are provided (user id and device id), and then every other optional parameter can be set manually. But I won't insist since this is bindings code, as long as you're happy with the API shape here.

bnjbvr (approved):
lgtm

src/libolm_migration.rs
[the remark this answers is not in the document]
richvdh: Worth noting that `pickled_account` is also required: it's just that we don't have an explicit check for it here because it's not strongly typed (it defaults to an empty string, which is an invalid value), so strictly speaking that's three required parameters.

I think the current arrangement is the least bad solution in terms of the js-side API.

src/libolm_migration.rs
```diff
@@ -0,0 +1,446 @@
...
+    #[wasm_bindgen(js_name = "backupVersion")]
+    pub backup_version: Option<String>,
+
+    /// The backup recovery key, as a base64-encoded string.
+    #[wasm_bindgen(js_name = "backupRecoveryKey")]
+    pub backup_recovery_key: Option<String>,
+
+    /// The private, base64-encoded, master cross-signing key.
+    #[wasm_bindgen(js_name = "crossSigningMasterKey")]
+    pub cross_signing_master_key: Option<String>,
```
BillCarsonFr, line 79: Maybe worth addind in the name of the field (or in the binding name) that it's the private part? Same for the 4 secrets?

src/libolm_migration.rs
```diff
@@ -0,0 +1,446 @@
...
+        .map(|k| BackupDecryptionKey::from_base64(k.as_str()))
+        .transpose()?;
+
+    let cross_signing = PrivateCrossSigningIdentity::empty(&user_id);
+    cross_signing
+        .import_secrets_unchecked(
+            data.cross_signing_master_key.as_deref(),
+            data.cross_signing_self_signing_key.as_deref(),
+            data.cross_signing_user_signing_key.as_deref(),
+        )
```
BillCarsonFr, line 161: Maybe we should `.zeroize()` the `data.cross_signing_*` at this point?

src/libolm_migration.rs
```diff
@@ -0,0 +1,446 @@
...
+    ///   store the vodozemac data.
+    #[wasm_bindgen(js_name = "migrateMegolmSessions")]
+    pub async fn migrate_megolm_sessions(
+        sessions: Array,
+        pickle_key: Uint8Array,
+        store_handle: &StoreHandle,
+    ) -> Result<JsValue, JsError> {
+        let pickle_key = pickle_key.to_vec();
+
+        let rust_sessions = sessions
```
BillCarsonFr, line 376: If one of the megolm session to import is for some reason malformed/corrupted, will the whole chunk fail to import?
Maybe it would be better to ignore "malformed" and log something but not interrupt all?

src/libolm_migration.rs
```diff
@@ -0,0 +1,446 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `sessions` - An `Array` of {@link PickledInboundGroupSession}s to import.
+    /// * `pickle_key` - The libolm pickle key that was used to pickle the megolm
+    ///   session objects.
+    /// * `store_handle` - A connection to the CryptoStore which will be used to
+    ///   store the vodozemac data.
+    #[wasm_bindgen(js_name = "migrateMegolmSessions")]
+    pub async fn migrate_megolm_sessions(
```
BillCarsonFr, line 369: This could be potentially a lot of sessions. Should we inform the caller to do that in chunks? Or have some progress reporting?

src/libolm_migration.rs
```diff
@@ -0,0 +1,446 @@
...
+        ))
+        .collect(),
+        room_id: libolm_session
+            .room_id
+            .clone()
+            .ok_or_else(|| JsError::new("Room ID not specified for megolm session"))?
+            .inner,
+        imported: libolm_session.imported,
+        backed_up: libolm_session.backed_up,
+        history_visibility: None,
```
BillCarsonFr, line 420: I wonder what `history_visibility` is used for InboundGroupSession. Is it for future support of shared history on invite?

src/libolm_migration.rs
```diff
@@ -0,0 +1,446 @@
...
+#[derive(Debug, Default)]
+pub struct Migration {}
+
+
+/// The base dataset that is important to migrate to the Rust SDK.
+///
+/// Can be imported into the rust store with {@link #migrateBaseData}.
+#[wasm_bindgen(getter_with_clone)]
+#[derive(Debug, Default)]
+pub struct BaseMigrationData {
```
BillCarsonFr, line 56: Shouldn't there be a list of tracked users here? or a `migrate_tracked_users` method?

BillCarsonFr (changes requested):
Just a few comments. I think that the migration of tracked users is missing?

src/libolm_migration.rs
richvdh: I'm not entirely convinced. This sounds like we could end up not migrating *any* sessions, and the user would just be confused.

Also: once the batch is successfully imported, we *delete* it from the legacy store. So, if we do fail to import any sessions, we'd end up completely throwing them away, which seems very dangerous.

I'd prefer to leave it and see if any errors actually arise in practice.

src/libolm_migration.rs
richvdh: The intention is that the application calls this with small enough batches that there is no need for an intermediate progress update.

src/libolm_migration.rs
BillCarsonFr: Not exacty related but we have seen sort of similar things when [importing from backup](https://github.com/element-hq/element-web/issues/26774), also when migrating tracked users (legacy is more lax on userIds).

Could we return a result with the succesfully migrated things? so that the caller would delete only these and handle the error on the application side?
I'll let you decide what's best

src/libolm_migration.rs
BillCarsonFr: Maybe document that?

src/libolm_migration.rs
richvdh: A very good question. I couldn't find anywhere that it was used.

It's initialised to `None` in lots of other places, so hopefully it's safe.

BillCarsonFr (approved):

src/libolm_migration.rs
richvdh: I wondered about that. I concluded it's not worthwhile, because even if we zeroize it on the Rust side, there may well still be copies kicking around in memory from the javascript side which we have no hope of zeroizing. (As https://github.com/veorq/cryptocoding#clean-memory-of-secret-data says: "there's virtually no way to reliably clean secret data in garbage-collected languages (such as ... Javascript)").

And we don't generally zeroize any other sensitive data that is passed through the wasm bindings.

src/libolm_migration.rs
richvdh: > Could we return a result with the succesfully migrated things? so that the caller would delete only these and handle the error on the application side?

Well, if we don't delete the broken sessions, then we'll just get stuck in a loop of repeatedly trying to migrate them. (There is no record of "which sessions have been migrated").)

I suppose one thing that we might think about is whether we can display an error to the user if migration goes wrong.

src/libolm_migration.rs
richvdh: As discussed elsewhere: I don't think there's any need to migrate the tracked users.

src/libolm_migration.rs
richvdh: 808a5736

PR #82, opened 2024-01-08
fix: Remove `try_array_to_vec` and `downcast`

### LINKED ISSUE #51: Downcast doesn't play nicely with minification
[This function](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/blob/207c54a41b647be1df65b9b584f20c0a07a0b42d/src/js.rs#L9) looks at constructor names to try to determine the right types. Specifically:
```rust
let constructor_name = Object::get_prototype_of(value).constructor().name();
```
This may return an unexpected constructor name when minification has been used, as classes are renamed to shorter strings e.g `UserId` becomes `NG`. This can then cause confusing errors:

```
Error: Expect an `UserId` instance, received `Ng` instead
```

Hywan:
This patch updates `wasm-bindgen` to 0.2.89. It allows to remove `try_array_to_vec` (which was fine, but no more necessary), and to remove `downcast` (which was a problem, see #51).

This patch should ideally be reviewed commit by commit.

---

* Closes #51


src/machine.rs
```diff
@@ -227,16 +225,19 @@ impl OlmMachine {
     /// "changed" notification for that user in the future.
     ///
     /// Users that were already in the list are unaffected.
+    ///
+    /// `users` is going to be dropped on the JS part after the function
+    /// returns. Be careful to not used `users` after that.
```
richvdh, line 230: "dropped" isn't really a term that means anything to the average JS developer. Could you perhaps phrase this in a way that will be easier to understand?

TBH I'm slightly confused myself. Is it the individual `UserID` objects, or the array, that will be invalidated? Or both?

Hywan: The individual `UserId` objects. That's how `wasm-bindgen` works. It doesn't make me super happy, but it is what it is.

I'll update the doc'.

src/identifiers.rs
```diff
@@ -53,6 +53,12 @@ impl UserId {
     pub fn to_string(&self) -> String {
         self.inner.as_str().to_owned()
     }
+
+    /// Clone this `UserId`.
```
richvdh, line 57:
```suggestion
    /// Create a clone of this `UserId`.
    ///
    /// This can be useful when passing a UserId instance to methods such as {@link OlmMachine.updateTrackedUsers}
    /// which destroy the instance.

File: src/machine.rs (hunk 1 above)
Comment (richvdh) [reply src/machine.rs:230]: Also: I feel like we should have a similar warning on all the other methods which consume a `Vec`, like `request_verification`.

**Review by richvdh — CHANGES_REQUESTED** [review]
lgtm generally, though please could the doc be updated as previously noted.

I don't suppose that wasm-bindgen supports `Vec<&identifiers::UserId>`, thus avoiding the drop problem?

Particularly given this is a not-very-obvious breaking change, could you add a note to the the CHANGELOG warning applications about this?

**Comment by Hywan** [comment]
> I don't suppose that wasm-bindgen supports Vec<&identifiers::UserId>, thus avoiding the drop problem?

It's not supported, indeed.

I've updated the docs.

File: CHANGELOG.md [hunk 3]
```diff
@@ -1,3 +1,17 @@
+# matrix-sdk-crypto-wasm unreleased
+
+-   Functions/methods that take a JavaScript `Array` as argument now delete the
+    array's items once their return, for example `Device::requestVerification`,
```
richvdh, line 4: AIUI, it's not that the entries are deleted from the array, but rather that the individual items are themselves invalidated so that they can no longer be used?

Also: I think it happens as soon as the method is called, rather than once it returns? In other words, I can't do this:

```javascript
    const user = new UserId("@user:domain");
    await Promise.all([
         machine.query_keys_for_users([user]),
         // `user` is used below before `query_keys_for_users` returns, but AIUI this will fail.
         machine.update_tracked_users([user]),
     ]);
```

```suggestion
-   Functions/methods that take a JavaScript `Array` as argument now invalidate the
    items within that array so that they cannot be re-used. Affected methods include `Device::requestVerification`,
```

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
...
+    `OlmMachine::updateTrackedUsers`, `OlmMachine::shareRoomKey`,
+    `OlmMachine::queryKeysForUsers`, `OlmMachine:getMissingSessions` and so on.
```
richvdh, line 6: I think we need to give a complete list of the affected methods, so that applications can audit their usages.

src/libolm_migration.rs
```diff
@@ -225,20 +224,21 @@ impl Migration {
     ///
     /// * `sessions` - An `Array` of {@link PickledSession}s to import.
     /// * `pickle_key` - The libolm pickle key that was used to pickle the olm
-    ///   session objects.
+    ///   session objects. Items inside `sessions` are going to be deleted on
+    ///   the JS part after the function returns.
```
richvdh, line 228: Not after the function returns.

```suggestion
    ///   session objects. Items inside `sessions` will be invalidated by this method.
```

src/identities.rs
```diff
@@ -60,9 +57,11 @@ impl OwnUserIdentity {
 
     /// Send a verification request to our other devices.
```
richvdh, line 58:
```suggestion
    /// Send a verification request to our other devices.
    ///
    /// Items inside `methods` will be invalidated by this method.
```

src/device.rs
```diff
@@ -28,10 +26,12 @@ impl Device {
     /// Returns a Promise for a 2-element array `[VerificationRequest,
     /// ToDeviceRequest]`.
```
richvdh, line 27:
```suggestion
    /// Items inside `methods` will be invalidated by this method.
    ///
    /// Returns a Promise for a 2-element array `[VerificationRequest,
    /// ToDeviceRequest]`.
```

src/identities.rs
```diff
@@ -164,13 +163,12 @@ impl UserIdentity {
         &self,
         room_id: &identifiers::RoomId,
         request_event_id: &identifiers::EventId,
-        methods: Option<Array>,
+        methods: Option<Vec<verification::VerificationMethod>>,
```
richvdh, line 166: as above, please document that this will consume entries in `methods`

src/identities.rs
```diff
@@ -188,10 +186,12 @@ impl UserIdentity {
     /// After the content has been sent out a VerificationRequest can be started
     /// with the `request_verification` method.
     #[wasm_bindgen(js_name = "verificationRequestContent")]
-    pub fn verification_request_content(&self, methods: Option<Array>) -> Result<Promise, JsError> {
+    pub fn verification_request_content(
+        &self,
+        methods: Option<Vec<verification::VerificationMethod>>,
```
richvdh, line 191: and here

src/machine.rs
```diff
@@ -227,16 +225,19 @@ impl OlmMachine {
     /// "changed" notification for that user in the future.
     ///
     /// Users that were already in the list are unaffected.
+    ///
+    /// Items inside `users` are going to be deleted on the JS part after the
+    /// function returns. Be careful to not used `users` after that.
```
richvdh, line 230:
```suggestion
    /// Items inside `users` will be invalidated by this method. Be careful to not use them after that.
```

richvdh: And similar corrections/clarifications to all the other methods in this file.

src/sync_events.rs
```diff
@@ -19,20 +19,14 @@ impl DeviceLists {
     ///
     /// `changed` and `left` must be an array of `UserId`.
     #[wasm_bindgen(constructor)]
-    pub fn new(changed: Option<Array>, left: Option<Array>) -> Result<DeviceLists, JsError> {
+    pub fn new(
+        changed: Option<Vec<identifiers::UserId>>,
+        left: Option<Vec<identifiers::UserId>>,
+    ) -> Result<DeviceLists, JsError> {
```
richvdh, line 25: once more, please document the invalidation behaviour

src/verification.rs
```diff
@@ -798,9 +797,9 @@ impl VerificationRequest {
         own_user_id: &UserId,
         own_device_id: &DeviceId,
         other_user_id: &UserId,
-        methods: Option<Array>,
+        methods: Option<Vec<VerificationMethod>>,
```
richvdh, line 800: and again

src/verification.rs
```diff
@@ -982,13 +981,17 @@ impl VerificationRequest {
     /// supports the given verification methods.
     ///
     /// `methods` represents the methods that we should advertise as
-    /// supported by us.
+    /// supported by us. Items inside `methods` are going to be deleted on the
+    /// JS part after the function returns.
```
richvdh, line 985: clarification needed here

richvdh (changes requested):
I'm sorry to be so picky about this, but this is such alien behaviour in javascript that I think it's really important to *accurately* document the hell out of it.

CHANGELOG.md
richvdh: actually, let's not bother with this. People can just go and look at the linked pull request.

Hywan:
I've rebased the patch. I've updated the documentation. I've also replaced a couple more `Array` by `Vec<T>` (in a return position). I've fixed the `CHANGELOG`. I believe we are good this time 🙂.

CHANGELOG.md
```diff
@@ -8,7 +8,22 @@
 
 **BREAKING CHANGES**
 
--   Rename `OlmMachine.init_from_store` introduced in v3.6.0 to `OlmMachine.initFromStore`.
+-   Rename `OlmMachine.init_from_store` introduced in v3.6.0 to
+    `OlmMachine.initFromStore`.
+    ([#84](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/84))
+
+**BREAKING CHANGES**
+
```
richvdh, line 16: we now have two BREAKING CHANGES headings. (This is why I prefer tools like towncrier than maintaining changelog files in PRs, but I digress.)

```suggestion
```

CHANGELOG.md
```diff
@@ -8,7 +8,22 @@
...
+-   Functions/methods that take a JavaScript `Array` as argument now invalidate the
+    items within that array so that they cannot be re-used as soon as
+    they are received by the functions/methods. See the patch for affected methods.
+    ([#82](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/82/))
+
+-   Update `wasm-bindgen` to 0.2.89. It allows to remove the `downcast` method.
+    It fixes [#51](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/51),
+    thus the resulting JavaScript code of `matrix-rust-sdk-crypto-wasm` can
+    be minified with no issue now.
+    ([#82](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/82/))
```
richvdh, line 26: This is not a breaking change.

```suggestion
**Other changes**

-   Update `wasm-bindgen` to 0.2.89. It allows to remove the `downcast` method.
    It fixes [#51](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/51),
    thus the resulting JavaScript code of `matrix-rust-sdk-crypto-wasm` can
    be minified with no issue now.
    ([#82](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/82/))
```

src/machine.rs
```diff
@@ -228,16 +226,19 @@ impl OlmMachine {
     /// "changed" notification for that user in the future.
     ///
     /// Users that were already in the list are unaffected.
+    ///
+    /// Items inside `users` will be invalidated by this method. Be careful to
+    /// not used `UserId`s after this method has been called.
```
richvdh, line 231:
```suggestion
    /// Items inside `users` will be invalidated by this method. Be careful not
    /// to use the `UserId`s after this method has been called.
```

src/machine.rs
```diff
@@ -637,21 +638,24 @@ impl OlmMachine {
     /// time is in flight for the same room, e.g. using a lock.
     ///
     /// Returns an array of `ToDeviceRequest`s.
+    ///
+    /// Items inside `users` will be invalidated by this method. Be careful to
+    /// not used `UserId`s after this method has been called.
```
richvdh, line 643:
```suggestion
    /// Items inside `users` will be invalidated by this method. Be careful not
    /// to use the `UserId`s after this method has been called.
```

src/machine.rs
```diff
@@ -675,12 +679,15 @@ impl OlmMachine {
     ///
     /// Returns a `KeysQueryRequest` object. The response of the request should
     /// be passed to the `OlmMachine` with the `mark_request_as_sent`.
+    ///
+    /// Items inside `users` will be invalidated by this method. Be careful to
+    /// not used `UserId`s after this method has been called.
```
richvdh, line 684:
```suggestion
    /// Items inside `users` will be invalidated by this method. Be careful not
    /// to use the `UserId`s after this method has been called.
```

src/machine.rs
```diff
@@ -713,13 +720,16 @@ impl OlmMachine {
     /// `users` represents the list of users that we should check if
     /// we lack a session with one of their devices. This can be an
     /// empty iterator when calling this method between sync requests.
+    ///
+    /// Items inside `users` will be invalidated by this method. Be careful to
+    /// not used `UserId`s after this method has been called.
```
richvdh, line 725:
```suggestion
    /// Items inside `users` will be invalidated by this method. Be careful not
    /// to use the `UserId`s after this method has been called.
```

src/machine.rs
```diff
@@ -1398,6 +1388,31 @@ impl OlmMachine {
...
 }
 
+impl OlmMachine {
+    /// Shared helper for `import_exported_room_keys` and `import_room_keys`.
+    ///
+    /// Wraps the progress listener in a Rust closure and runs
+    /// `Store::import_exported_room_keys`
+    ///
+    /// Items inside `exported_room_keys` will be invalidated by this method. Be
+    /// careful to not used `ExportedRoomKey`s after this method has been
```
richvdh, line 1398:
```suggestion
    /// careful not to use the `ExportedRoomKey`s after this method has been
```

src/sync_events.rs
```diff
@@ -18,21 +17,18 @@ impl DeviceLists {
     /// Create an empty `DeviceLists`.
     ///
     /// `changed` and `left` must be an array of `UserId`.
+    ///
+    /// Items inside `changed` and `left` will be invalidated by this method. Be
+    /// careful to not used `UserId`s after this method has been called.
```
richvdh, line 22:
```suggestion
    /// Items inside `changed` and `left` will be invalidated by this method. Be
    /// careful not to use the `UserId`s after this method has been called.
```

src/verification.rs
```diff
@@ -982,13 +948,19 @@ impl VerificationRequest {
...
+    ///
+    /// Items inside `methods` will be invalidated by this method.
```
richvdh, line 954: We're saying the same thing twice here:

```suggestion
    /// supported by us.
    ///
    /// Items inside `methods` will be invalidated by this method.
```

richvdh (approved):
Thank you, looks great apart from a few more documentation nits!

PR #85, opened 2024-01-12
don't die when we have invalid keys in backup

uhoreg:
Counts the invalid keys from backup and reports them to the callback, rather than returning an error immediately and stopping import.  Works with JS-SDK PR https://github.com/matrix-org/matrix-js-sdk/pull/4006

Fixes https://github.com/element-hq/element-web/issues/26774

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # UNRELEASED
 
 -   Report failures to callback when importing backed-up room keys. The
-    callback is now called with a third argument, giving the number of invalid
+    `progress_listener` callback in the `OlmMachine.importBackedUpRoomKeys`
+    function is now called with a third argument, giving the number of invalid
     room keys.
     ([#85](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/85))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # UNRELEASED
 
--   Report failures to callback when importing backed-up room keys.  The
+-   Report failures to callback when importing backed-up room keys. The
     callback is now called with a third argument, giving the number of invalid
     room keys.
     ([#85](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/85))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 -   Report failures to callback when importing backed-up room keys.  The
     callback is now called with a third argument, giving the number of invalid
     room keys.
+    ([#85](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/85))
 
 **BREAKING CHANGES**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # UNRELEASED
 
+-   Report failures to callback when importing backed-up room keys.  The
+    callback is now called with a third argument, giving the number of invalid
+    room keys.
+
 **BREAKING CHANGES**
 
 -   Rename `OlmMachine.init_from_store` introduced in v3.6.0 to `OlmMachine.initFromStore`.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # UNRELEASED
 
+-   Report failures to callback when importing backed-up room keys. The
+    callback is now called with a third argument, giving the number of invalid
+    room keys.
```
richvdh, line 5: It would be helpful to mention which API is affected here. `OlmMachine.importBackedUpRoomKeys` I think?

src/machine.rs
```diff
@@ -1022,7 +1028,12 @@ impl OlmMachine {
                 .import_backed_up_room_keys(keys, |progress, total| {
                     if let Some(callback) = &progress_listener {
                         callback
-                            .call2(&JsValue::NULL, &JsValue::from(progress), &JsValue::from(total))
+                            .call3(
```
richvdh, line 1031: documentation on this method needs an update.

src/machine.rs
```diff
@@ -1009,10 +1010,15 @@ impl OlmMachine {
...
-
-                keys.entry(room_id.clone()).or_default().insert(session_id.into(), key);
+                serde_wasm_bindgen::from_value::<BackedUpRoomKey>(room_room_keys_entry.get(1))
+                    .map_or_else(
+                        |_| {
+                            failures = failures + 1;
+                        },
+                        |key| {
+                            keys.entry(room_id.clone()).or_default().insert(session_id.into(), key);
+                        },
```
richvdh, line 1020: personally I'd find a `match` statement easier to read here, but ymmv

src/machine.rs
```diff
@@ -998,6 +998,7 @@ impl OlmMachine {
 
         // convert the js-side data into rust data
         let mut keys: BTreeMap<_, BTreeMap<_, _>> = BTreeMap::new();
+        let mut failures: usize = 0;
```
richvdh, line 1001: I don't know if this is more idiomatic:

```suggestion
        let mut failures = 0usize;
```

richvdh (approved):
Functionality looks good to me, but documentation needs an update and it might be worth a quick look from a Rust expert (cc @Hywan / @bnjbvr) to help with the idiomatic questions

src/machine.rs
bnjbvr: Indeed, since the error is ignored anyways, even an if/else would read better: `if let Ok(key) = serde_wasm_bindgen::... { keys.entry(...)... } else { failures += 1; }`

src/machine.rs
bnjbvr: Can't the compiler infer it should be a usize? If not, can we use `let mut failures = 0` here (assuming i32::max_value() is sufficient for the max number of failures) and then the one user can be replaced with `JsValue::from(failures as usize)`?

bnjbvr (commented):
Always here to answer the ~~bat~~ crab-signal :crab:

uhoreg:
I've made the requested changes. Someone else will have to merge because I don't have permissions in this repo

src/machine.rs
```diff
@@ -1022,7 +1028,12 @@ impl OlmMachine {
                 .import_backed_up_room_keys(keys, |progress, total| {
                     if let Some(callback) = &progress_listener {
                         callback
-                            .call2(&JsValue::NULL, &JsValue::from(progress), &JsValue::from(total))
+                            .call3(
+                                &JsValue::NULL,
+                                &JsValue::from(progress),
+                                &JsValue::from(total + failures),
```
bnjbvr, line 1034: I don't understand this line. Does the total not include failures too? If so, should the `total` variable be renamed `successes` maybe?

uhoreg: You're right, it's not very clear.  I've added a comment and renamed the variable, so it should be clearer now.

andybalaam:
@uhoreg want me to hit merge?

PR #86, opened 2024-01-18
Update matrix-rust-sdk version

richvdh:
... and disable the state store in `matrix-sdk-indexeddb`. We don't use the state store, so can save some build time by excluding it (since this also means we no longer pull in `matrix-sdk-base`).


bnjbvr (approved):
Hurrah to shipping less code \o/ Does this impact the binary size, out of curiosity?

richvdh:
> Hurrah to shipping less code \o/ Does this impact the binary size, out of curiosity?

Not hugely: the wasm is 5.3M, where it was 5.5M before. A 4% improvement isn't nothing, I guess.

PR #87, opened 2024-01-19
v4.0.0

richvdh:


bnjbvr (approved):
Yay! (no need to run yarn or something so the lock file references the new version of the library?)

richvdh:
> (no need to run yarn or something so the lock file references the new version of the library?)

The (javascript) dependencies aren't changing here, so there's nothing to update in `yarn.lock`.

PR #88, opened 2024-01-19
add callback for getting device changes

uhoreg:
part of fix for https://github.com/element-hq/element-web/issues/26263


src/machine.rs
```diff
@@ -1267,6 +1271,25 @@ impl OlmMachine {
...
+    /// Register a callback which will be called whenever there is an update to
+    /// a device.
+    ///
+    /// `callback` should be a function that takes a single argument (an array
+    /// of user IDs) and returns a Promise.
+    #[wasm_bindgen(js_name = "registerDevicesUpdatedCallback")]
+    pub async fn register_devices_updated_callback(&self, callback: Function) {
+        let stream = self.inner.store().identities_stream_raw();
+
+        // fire up a promise chain which will call `cb` on each result from the stream
```
Hywan, line 1283:
```suggestion
        // fire up a promise chain which will call `callback` on each result from the stream
```

Hywan (approved):
The Rust code looks good to me. Thanks!

uhoreg:
Can someone hit the "Merge" button for me?

CHANGELOG.md
```diff
@@ -25,6 +25,10 @@
     room keys.
     ([#85](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/85))
 
+-   Add `OlmMachine.registerDevicesUpdatedCallback` to notify when devices have
+    been update.
+    ([#88](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/88))
+
```
richvdh, line 31: This won't be part of v4.0.0. Please could you move it to an "UNRELEASED" section

src/machine.rs
```diff
@@ -1267,6 +1271,26 @@ impl OlmMachine {
         });
     }
 
+    /// Register a callback which will be called whenever there is an update to
+    /// a device.
+    ///
+    /// `callback` should be a function that takes a single argument (an array
+    /// of user IDs) and returns a Promise.
```
richvdh, line 1278: Please clarify if the user IDs are `String`s or `UserID`s

src/machine.rs
```diff
@@ -1447,6 +1471,34 @@ async fn send_user_identities_to_callback(
...
+    // get the user IDs of all the devices that have changed
+    let updated_chain = device_updates
+        .new
+        .into_iter()
+        .chain(device_updates.changed.into_iter())
+        .chain(device_updates.deleted.into_iter());
+    // put them in a set to make them unique
+    let updated_users: HashSet<String> =
+        HashSet::from_iter(updated_chain.map(|device| device.user_id().to_string()));
+    let updated_users_vec = Vec::from_iter(updated_users.iter());
```
richvdh, line 1489: I think this means the user IDs are passed as `String`s? Which seems inconsistent with other parts of the API, like `register_user_identity_updated_callback` ?

richvdh (changes requested):

src/machine.rs
uhoreg: Yes, it is.  It's because HashSet (and BTreeSet) didn't like storing `identifiers::UserId`, and it seemed silly to convert it to a string for HashSet, and then convert it to a `UserId`, only for the JavaScript to convert it back to a string.

richvdh (commented):
We should really be in the habit of having tests. I think a test shouldn't be too difficult to write: there are various tests in the test suite already which mock up a response to a `/keys/query` request, for example?

uhoreg:
test added

richvdh (approved):
lgtm, thank you!

PR #90, opened 2024-01-22
Update documentation added in #82

richvdh:
#82 updated the documentation on some of the exported methods. This PR cleans up a few bits. Principally: `VerificationMethod` is an int, so there is no "owner", and it is not invalidated by passing it into a Rust method.

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -1392,11 +1392,7 @@ impl OlmMachine {
     /// Shared helper for `import_exported_room_keys` and `import_room_keys`.
     ///
     /// Wraps the progress listener in a Rust closure and runs
-    /// `Store::import_exported_room_keys`
-    ///
-    /// Items inside `exported_room_keys` will be invalidated by this method. Be
-    /// careful not to use the `ExportedRoomKey`s after this method has been
-    /// called.
+    /// `Store::import_exported_room_keys`.
     async fn import_exported_room_keys_helper(
         inner: &matrix_sdk_crypto::OlmMachine,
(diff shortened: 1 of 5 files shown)


src/machine.rs
```diff
@@ -1392,11 +1392,7 @@ impl OlmMachine {
     /// Shared helper for `import_exported_room_keys` and `import_room_keys`.
     ///
     /// Wraps the progress listener in a Rust closure and runs
-    /// `Store::import_exported_room_keys`
-    ///
-    /// Items inside `exported_room_keys` will be invalidated by this method. Be
-    /// careful not to use the `ExportedRoomKey`s after this method has been
-    /// called.
```
richvdh, line 1399: this method isn't actually called by javascript. So this is just standard rust semantics, which don't need explicit documentation.

src/verification.rs
```diff
@@ -849,6 +849,11 @@ impl VerificationRequest {
     ///
     /// Will be present only if the other side requested the
     /// verification or if we’re in the ready state.
+    ///
+    /// # Returns
+    ///
+    /// `undefined` if we do not yet know the supported methods; otherwise, an
+    /// array of `VerificationMethod`s.
```
richvdh, line 856: this documentation was removed in #82 (https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/82/files#diff-cf8f469cee769b4b3555833df9943f95a94dd55a4c6c5c8b8c7419e34fc8b5d8L879).

andybalaam (approved):
Looks good, thanks!

PR #92, opened 2024-01-23
add unknown to EncryptionAlgorithm enum

uhoreg:
Related: https://github.com/element-hq/element-web/issues/26610, which will be fixed by this once matrix-js-sdk is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 -   `PickledInboundGroupSession.sender_signing_key` is now optional.
     ([#89](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/89))
 
+-   Add `Unknown` to `EncryptionAlgorithm`, representing unsupported algorithms
+    coming from matrix-sdk-crypto's `EventEncryptionAlgorithm`.
+    ([#92](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/92))
+
 # matrix-sdk-crypto-wasm v4.0.0
 
 -   Properly encode missing and `Duration` parameters in requests.
--- a/src/encryption.rs
+++ b/src/encryption.rs
@@ -85,8 +85,9 @@ pub enum EncryptionAlgorithm {
     /// Megolm version 1 using AES-256 and SHA-256.
     MegolmV1AesSha2,
 
-    /// Unknown algorithm (JavaScript should ignore this value if it is
-    /// received, and should never send this value to Rust)
+    /// Unsupported algorithm.
+    ///
+    /// Applications should ignore this value if it is received, and should never set it.
     Unknown,
 }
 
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 -   `PickledInboundGroupSession.sender_signing_key` is now optional.
     ([#89](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/89))
 
+-   Add `Unknown` to `EncryptionAlgorithm`, representing unsupported algorithms
```
uhoreg, line 6: Does this count as a breaking change?

src/encryption.rs
```diff
@@ -77,13 +77,17 @@ impl From<&EncryptionSettings> for matrix_sdk_crypto::olm::EncryptionSettings {
...
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum EncryptionAlgorithm {
     /// Olm version 1 using Curve25519, AES-256, and SHA-256.
     OlmV1Curve25519AesSha2,
 
     /// Megolm version 1 using AES-256 and SHA-256.
     MegolmV1AesSha2,
+
+    /// Unknown algorithm (JavaScript should ignore this value if it is
+    /// received, and should never send this value to Rust)
```
richvdh, line 89: From the point of view of a reader of this documentation, the fact that the library happens to be implemented in Rust is an implementation detail.

Also: it's not that we don't know what algorithm the device uses; rather it's that we don't support it.

```suggestion
    /// Unsupported algorithm.
    ///
    /// Applications should ignore this value if it is received, and should never set it.
```

CHANGELOG.md
richvdh: Kinda. The js-sdk code to handle this is going to be incorrect, for instance, since `Unknown` != `Olm`.

```ts
    rustAlgorithms.forEach((algorithm) => {
        switch (algorithm) {
            case RustSdkCryptoJs.EncryptionAlgorithm.MegolmV1AesSha2:
                algorithms.add("m.megolm.v1.aes-sha2");
                break;
            case RustSdkCryptoJs.EncryptionAlgorithm.OlmV1Curve25519AesSha2:
            default:
                algorithms.add("m.olm.v1.curve25519-aes-sha2");
                break;
        }
    });
```

... But I'm not really sure it's worth a major version bump. I'd treat it as a normal change.

Please put it at the top of the changelog though (most recent changes in the changelog are first, obviously), and resolve the conflict.

richvdh (changes requested):

richvdh:
> fixes https://github.com/element-hq/element-web/issues/26774

no, it doesn't. Not until the js-sdk is updated to use this change.

uhoreg:
I linked to the wrong issue in the PR description.  Fixed now.

richvdh:
> I linked to the wrong issue in the PR description. Fixed now.

Ok, but please avoid the magic github words "will fix <issue>", which means that GH will auto-close that issue when the PR is merged, irrespective of the fact we have many more steps to do before that happens. Now fixed.

richvdh (approved):
thanks!

uhoreg:
> > I linked to the wrong issue in the PR description. Fixed now.
> 
> Ok, but please avoid the magic github words "will fix ", which means that GH will auto-close that issue when the PR is merged, irrespective of the fact we have many more steps to do before that happens. Now fixed.

Oh, I changed "fixes" to "will fix" because I thought that would be safe. :-/

PR #93, opened 2024-01-23
Fix changelog for 4.0.1

richvdh:
PR #72 landed in 4.0.1, not 4.0.0.


bnjbvr (approved):
As suggested, maybe we could add an `# Unreleased` section at the top so people who just add to the top paragraph wouldn't be wrong?

richvdh:
> As suggested, maybe we could add an # Unreleased section at the top so people who just add to the top paragraph wouldn't be wrong?

Yeah. Added. I'll update the release instructions too at some point.

PR #95, opened 2024-01-25
Support for saving and loading room settings

richvdh:
Part of https://github.com/element-hq/element-web/issues/26108

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -32,7 +32,7 @@ use crate::{
     store,
     store::{RoomKeyInfo, StoreHandle},
     sync_events,
-    types::{self, RoomKeyImportResult, SignatureVerification},
+    types::{self, RoomKeyImportResult, RoomSettings, SignatureVerification},
     verification, vodozemac,
 };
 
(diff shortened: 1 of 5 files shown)


src/machine.rs
```diff
@@ -1379,6 +1379,42 @@ impl OlmMachine {
...
+    ) -> Result<JsValue, JsError> {
+        let result = self.inner.room_settings(&room_id.inner).await?;
+        Ok(result.map(|settings| RoomSettings::from(settings)).into())
+    }
+
+    /// Store encryption settings for the given room.
+    ///
+    /// This method checks if the new settings are "safe" -- ie, that they do
+    /// not represent a downgrade in encryption security from any previous
+    /// settings. Attempts to downgrade security will result in an error.
```
andybalaam, line 1403: I don't see these checks in this code. Are they in the code we call into?

andybalaam (approved):
Looks great, thanks!

src/machine.rs
richvdh: Yup.

PR #101, opened 2024-02-06
Add support for sending encrypted to_device

BillCarsonFr:
Depends on https://github.com/matrix-org/matrix-rust-sdk/pull/3091

Add bindings to support sending custom encrypted to device event.
see https://github.com/matrix-org/matrix-js-sdk/issues/3304


tests/device.test.js
```diff
@@ -146,6 +146,140 @@ describe(OlmMachine.name, () => {
     });
 });
 
+describe("Send to device", () => {
+    const userId1 = new UserId("@alice:example.org");
+    const deviceId1 = new DeviceId("alice_device");
+
+    const userId2 = new UserId("@bob:example.org");
+    const deviceId2 = new DeviceId("bob_device");
+
```
richvdh, line 155: these appear unused?

tests/device.test.js
```diff
@@ -146,6 +146,140 @@ describe(OlmMachine.name, () => {
...
+            const marked = await m.markRequestAsSent("bar", RequestType.KeysClaim, hypotheticalResponse);
+        }
+
+        // Pick the device we want to encrypt to.
+        const device2 = await m.getDevice(new UserId("@example:localhost"), new DeviceId("AFGUOBTZWM"));
+
+        const content = {
+            body: "Hello, World!",
+        };
+        const type = "some.custon.event.type";
```
richvdh, line 272:
```suggestion
        const type = "some.custom.event.type";
```

tests/device.test.js
```diff
@@ -146,6 +146,140 @@ describe(OlmMachine.name, () => {
...
+                                "g4ED07Fnqf3GzVWNN1pZ0IFrPQVdqQf+PYoJNH4eE0s",
+                        },
+                        signatures: {
+                            "@example:localhost": {
+                                "ed25519:n2lpJGx0LiKnuNE1IucZP3QExrD4SeRP0veBHPe3XUU":
+                                    "nKQu8alQKDefNbZz9luYPcNj+Z+ouQSot4fU/A23ELl1xrI06QVBku/SmDx0sIW1ytso0Cqwy1a+3PzCa1XABg",
+                            },
+                        },
+                    },
+                },
```
richvdh, line 238: for simplicity, could we omit the x-signing keys (and the x-signing signature above) from this response?

richvdh (changes requested):

src/device.rs
```diff
@@ -51,6 +54,38 @@ impl Device {
...
+    /// available for the target device. This can be done by calling the
+    /// [`get_missing_sessions()`](OlmMachine::get_missing_sessions)
+    ///
+    /// The caller is responsible for sending the encrypted
+    /// event to the target device. If multiple messages are
+    /// encrypted using this method they should be sent in the same order as
+    /// they are encrypted.
+    ///
+    /// # Returns
+    /// Returns a Promise for a `json` string of the encrypted event.
```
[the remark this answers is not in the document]
BillCarsonFr: Finally found the way to do it (needed to use `Serializer::json_compatible()`)

tests/device.test.js
BillCarsonFr: yes good catch

tests/device.test.js
```diff
@@ -146,6 +146,140 @@ describe(OlmMachine.name, () => {
...
 
+describe("Send to device", () => {
+    const userId1 = new UserId("@alice:example.org");
+    const deviceId1 = new DeviceId("alice_device");
+
+    const userId2 = new UserId("@bob:example.org");
+    const deviceId2 = new DeviceId("bob_device");
+
+    function machine(newUser, newDevice) {
+        return OlmMachine.initialize(newUser || userId1, newDevice || deviceId1);
```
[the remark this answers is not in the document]
BillCarsonFr: removed

tests/device.test.js
```diff
@@ -146,6 +146,140 @@ describe(OlmMachine.name, () => {
...
+    }
+
+    it("can send a to-device message", async () => {
+        // Olm machine.
+        const m = await machine(userId1, deviceId1);
+
+        // Make m aware of another device, and get some OTK to be able to establish a session.
+        {
+            // derived from https://github.com/matrix-org/matrix-rust-sdk/blob/7f49618d350fab66b7e1dc4eaf64ec25ceafd658/benchmarks/benches/crypto_bench/keys_query.json
+            const hypotheticalResponse = JSON.stringify({
```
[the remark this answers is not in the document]
BillCarsonFr: done

src/device.rs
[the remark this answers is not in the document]
richvdh: OH! Sorry. When I asked this, I hadn't realised that `encrypt_event_raw` returns JSON. I don't think it's correct that we then deserialize that object back into a rust-side `ToDeviceEncryptedEventContent`, and then re-serialize *that* back to a js object.

Since we have JSON, let's just return it and deal with it on the JS side. Sorry!

richvdh (changes requested):
Looks good except for the complicated serialization stuff :/

src/device.rs
```diff
@@ -1,13 +1,18 @@
...
+use serde_wasm_bindgen::Serializer;
 use wasm_bindgen::prelude::*;
 
 use crate::{
     encryption::EncryptionAlgorithm,
     future::future_to_promise,
     identifiers::{self, DeviceId, UserId},
-    impl_from_to_inner, requests, types, verification, vodozemac,
+    impl_from_to_inner,
+    requests::{self},
```
Hywan, line 14:
```suggestion
    requests,
```

src/device.rs
```diff
@@ -51,6 +56,40 @@ impl Device {
         }))
     }
 
+    /// Encrypt a to-device message to be sent to this device, using Olm
+    /// encryption.
+    ///
+    /// Prior to calling this method you must ensure that an olm session is
```
Hywan, line 62:
```suggestion
    /// Prior to calling this method you must ensure that an Olm session is
```

src/device.rs
```diff
@@ -51,6 +56,40 @@ impl Device {
...
+    /// available for the target device. This can be done by calling
+    /// {@link OlmMachine.getMissingSessions}.
+    ///
+    /// The caller is responsible for sending the encrypted
+    /// event to the target device. If multiple messages are
+    /// encrypted for the same device using this method they should be sent in
+    /// the same order as they are encrypted.
+    ///
+    /// # Returns
+    /// Returns a Promise for a js object of the encrypted event.
```
Hywan, line 72:
```suggestion
    /// Returns a Promise for a JS object of the encrypted event.
```

src/device.rs
```diff
@@ -51,6 +56,40 @@ impl Device {
...
+    /// Can be used to create the payload for a `/sendToDevice` API.
+    #[wasm_bindgen(js_name = "encryptToDeviceEvent")]
+    pub fn encrypt_to_device_event(
+        &self,
+        event_type: String,
+        content: JsValue,
+    ) -> Result<Promise, JsError> {
+        let me = self.inner.clone();
+        let content: Value = serde_wasm_bindgen::from_value(content)?;
+        let event_type = event_type.clone();
```
Hywan, line 82: This line can be removed. No need to clone the event type here, you already own it.

Hywan (dismissed):

src/device.rs
richvdh:
```suggestion
```

src/device.rs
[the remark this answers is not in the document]
richvdh: [Arguably it seems a shame that `encrypt_event_raw` serializes its response, but that's a sidebar]

richvdh (approved):

PR #104, opened 2024-02-12
Add bindings for dehydrated devices

uhoreg:
Depends on https://github.com/matrix-org/matrix-rust-sdk/pull/3164

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,13 @@
 # UNRELEASED
 
+-   Add `OlmMachine.dehydratedDevices()` and `DehydratedDevices` class to
+    support dehydrated devices.
+    ([#104](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/104))
+
 -   Fix a problem when using matrix-sdk-crypto-wasm in a webapp running
     in the webpack dev server; when rebuilding, the server would throw an
     error.
     ([#109](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/109))
-
--   Add `OlmMachine.dehydratedDevices()` and `DehydratedDevices` class to
-    support dehydrated devices.
-    ([#104](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/104))
 
 # matrix-sdk-crypto-wasm v4.7.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # UNRELEASED
 
+-   Add `OlmMachine.dehydratedDevices()` and `DehydratedDevices` class to
+    support dehydrated devices.
+    ([#104](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/104))
+
 # matrix-sdk-crypto-wasm v4.4.0
 
 -   Update dependencies, including matrix-rust-sdk to
--- a/src/dehydrated_devices.rs
+++ b/src/dehydrated_devices.rs
@@ -1,0 +1,117 @@
+//! Dehydrated devices
+//!
+//! WASM wrapper for `matrix_sdk_crypto::dehydrated_devices`.
+
+use js_sys::{Array, JsString, Uint8Array};
+use matrix_sdk_crypto::dehydrated_devices;
+use wasm_bindgen::prelude::*;
+
+use crate::{identifiers::DeviceId, requests::PutDehydratedDeviceRequest, store::RoomKeyInfo};
+
+#[wasm_bindgen]
+#[derive(Debug)]
+/// Struct collecting methods to create and rehydrate dehydrated devices.
+pub struct DehydratedDevices {
+    inner: dehydrated_devices::DehydratedDevices,
+}
+
+impl From<dehydrated_devices::DehydratedDevices> for DehydratedDevices {
+    fn from(value: dehydrated_devices::DehydratedDevices) -> Self {
+        Self { inner: value }
+    }
+}
+
+#[wasm_bindgen]
+impl DehydratedDevices {
+    /// Create a new [`DehydratedDevice`] which can be uploaded to the server.
+    #[wasm_bindgen]
+    pub async fn create(&self) -> Result<DehydratedDevice, JsError> {
+        Ok(self.inner.create().await?.into())
+    }
+
+    /// Rehydrate a dehydrated device.
+    #[wasm_bindgen]
+    pub async fn rehydrate(
+        &self,
+        pickle_key: &Uint8Array,
+        device_id: &DeviceId,
+        device_data: &str,
+    ) -> Result<RehydratedDevice, JsError> {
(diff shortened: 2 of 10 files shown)


src/requests.rs
```diff
@@ -597,6 +600,69 @@ impl TryFrom<&OriginalUploadSigningKeysRequest> for UploadSigningKeysRequest {
...
+#[wasm_bindgen]
+impl PutDehydratedDeviceRequest {
+    /// Create a new `PutDehydratedDeviceRequest`
+    #[wasm_bindgen(constructor)]
+    pub fn new(body: JsString) -> PutDehydratedDeviceRequest {
+        Self { body }
+    }
+}
+
+impl TryFrom<OriginalPutDehydratedDeviceRequest> for PutDehydratedDeviceRequest {
```
BillCarsonFr, line 623: I think there is a macro to do that.
Like `request!(KeysBackupRequest from OriginalKeysBackupRequest extracts version: string and groups rooms);`

tests/dehydrated_devices.test.ts
```diff
@@ -0,0 +1,30 @@
...
+        const req = await device.keysForUpload("Dehydrated device", key);
+        const body = JSON.parse(req.body);
+
+        const machine2 = await OlmMachine.initialize(new UserId("@alice:example.org"), new DeviceId("HIJKLMN"));
+        const dehydratedDevices2 = machine2.dehydratedDevices();
+        const rehydratedDevice = await dehydratedDevices2.rehydrate(
+            key,
+            new DeviceId(body.device_id),
+            JSON.stringify(body.device_data),
+        );
```
BillCarsonFr, line 28: Can we make some basic expect here?

src/dehydrated_devices.rs
```diff
@@ -0,0 +1,117 @@
...
+use js_sys::{Array, JsString, Uint8Array};
+use matrix_sdk_crypto::dehydrated_devices;
+use wasm_bindgen::prelude::*;
+
+use crate::{identifiers::DeviceId, requests::PutDehydratedDeviceRequest, store::RoomKeyInfo};
+
+#[wasm_bindgen]
+#[derive(Debug)]
+/// Struct collecting methods to create and rehydrate dehydrated devices.
+pub struct DehydratedDevices {
```
BillCarsonFr, line 14: Was a bit surprised by the plural here, but it's the same as in the SDK.
It's more a Factory? or a Manager?

BillCarsonFr (approved):
Just a few comments

src/requests.rs
uhoreg: I think that the macro only works for requests that have an ID (which is why `UploadSigningKeys` is also not done using the macro).

src/dehydrated_devices.rs
uhoreg: Sort of a manager, but not quite.  Yeah, I'm not super happy with the naming, but it's what the SDK uses.

tests/dehydrated_devices.test.ts
uhoreg: we're just checking that it doesn't raise an exception, which means that the device was rehydrated successfully

uhoreg: The API doesn't give us the device keys for the rehydrated device.

uhoreg: I'll see if I can figure out how to encrypt to the dehydrated device, so we can test that the rehydrated device can decrypt.

uhoreg: Test updated

src/libolm_migration.rs
```diff
@@ -141,9 +141,11 @@ async fn migrate_base_data_to_store(
             // device keys to the server (it does it every time the stack is started). For safety,
             // let's assume it hasn't happened yet.
             shared: false,
+            dehydrated: false,
             // Assume we have 50 keys on the server, until we get a sync that says fewer.
             uploaded_signed_key_count: 50,
             creation_local_time: MilliSecondsSinceUnixEpoch::now(),
+            fallback_key_creation_timestamp: None,
```
uhoreg, line 148: This change is to make it work with latest version of matrix-sdk-crypto

src/requests.rs
```diff
@@ -512,10 +515,6 @@ pub fn outgoing_request_to_js_value(
         OutgoingRequests::RoomMessage(request) => {
             JsValue::from(RoomMessageRequest::try_from((request_id, request))?)
         }
-
-        OutgoingRequests::KeysBackup(request) => {
-            JsValue::from(KeysBackupRequest::try_from((request_id, request))?)
-        }
```
uhoreg, line 518: `OutgoingRequests` was deleted in latest version of `matrix-sdk-crypto` (https://github.com/matrix-org/matrix-rust-sdk/commit/3ccd2e9f8fb2763ea8be46fbebf79d8807d2851f)

uhoreg:
Test failure is because the latest matrix-sdk-crypto includes time-based fallback key rotation, which uses SystemTime to calculate the expiry, which doesn't work with wasm.  https://github.com/matrix-org/matrix-rust-sdk/pull/3249 fixes this.  So once that's merged, the `Cargo.lock` here will need to be updated again.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
     error.
     ([#109](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/109))
 
+-   Add `OlmMachine.dehydratedDevices()` and `DehydratedDevices` class to
+    support dehydrated devices.
+    ([#104](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/104))
+
```
richvdh, line 11: please can we put more recent changelog entries at the top?

src/requests.rs
```diff
@@ -597,6 +596,69 @@ impl TryFrom<&OriginalUploadSigningKeysRequest> for UploadSigningKeysRequest {
     }
 }
 
+#[wasm_bindgen(getter_with_clone)]
+#[derive(Debug)]
+/// Upload a dehydrated device to the server.
```
richvdh, line 601:
```suggestion
/// A request that will upload a dehydrated device to the server.
```

src/requests.rs
richvdh: It feels weird to me that this has to happen in the bindings, macro or not. Surely turning the request object into the relevant JSON is something that all applications will need to do, so it should be in the main SDK?

I appreciate it's following a pattern, so this is maybe something to fix another time, but still, it seems wrong.

richvdh (approved):
A couple of nits, but lgtm otherwise

CHANGELOG.md
uhoreg: Oh, I purposely put it at the bottom, because I mis-remembered the convention. :facepalm:

richvdh: yeahh... my thinking goes, if new *versions* go at the top, then by extension, so should individual entries, so that the newest stuff is always at the top.

PR #105, opened 2024-02-20
Use export_room_keys_stream to reduce export memory usage

andybalaam:
Applies on top of https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/106

Make use of the new `export_room_keys_stream` method from https://github.com/matrix-org/matrix-rust-sdk/pull/3144 to avoid one copy of the keys in the export process, by directly serialising to a JSON string, without taking the intermediate step into a `Vec<ExportedRoomKey>`.

Part of https://github.com/element-hq/element-web/issues/26681

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -3,10 +3,11 @@
 use std::{
     collections::{BTreeMap, HashSet},
     ops::Deref,
+    pin::{pin, Pin},
     time::Duration,
 };
 
-use futures_util::{pin_mut, StreamExt};
+use futures_util::{pin_mut, Stream, StreamExt};
 use js_sys::{Array, Function, JsString, Map, Promise, Set};
 use matrix_sdk_common::ruma::{
     self, events::secret::request::SecretName, serde::Raw, DeviceKeyAlgorithm, OwnedDeviceId,
@@ -19,8 +20,8 @@ use matrix_sdk_crypto::{
     types::RoomKeyBackupInfo,
     CryptoStoreError, EncryptionSyncChanges, GossippedSecret,
 };
+use serde::{ser::SerializeSeq, Serialize, Serializer};
 use serde_json::json;
-use serde_wasm_bindgen;
 use tracing::warn;
 use wasm_bindgen::{convert::TryFromJsValue, prelude::*};
 use wasm_bindgen_futures::{spawn_local, JsFuture};
@@ -883,15 +884,18 @@ impl OlmMachine {
     /// `predicate` is a closure that will be called for every known
     /// `InboundGroupSession`, which represents a room key. If the closure
     /// returns `true`, the `InboundGroupSession` will be included in the
-    /// export, otherwise it won't.
+    /// export; otherwise it won't.
+    ///
+    /// Returns a Promise containing a Result containing a String which is a
+    /// JSON-encoded array of ExportedRoomKey objects.
     #[wasm_bindgen(js_name = "exportRoomKeys")]
     pub fn export_room_keys(&self, predicate: Function) -> Promise {
         let me = self.inner.clone();
 
         future_to_promise(async move {
-            Ok(serde_json::to_string(
-                &me.store()
-                    .export_room_keys(|session| {
+            stream_to_json_array(pin!(
+                me.store()
+                    .export_room_keys_stream(|session| {
                         let session = session.clone();
 
                         predicate
@@ -901,7 +905,8 @@ impl OlmMachine {
                             .unwrap_or(false)
                     })
                     .await?,
-            )?)
+            ))
+            .await
         })
     }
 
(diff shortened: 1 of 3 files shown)


src/machine.rs
```diff
@@ -19,8 +20,9 @@ use matrix_sdk_crypto::{
     types::RoomKeyBackupInfo,
     CryptoStoreError, EncryptionSyncChanges, GossippedSecret,
 };
+use serde::{ser::SerializeSeq, Serialize, Serializer};
 use serde_json::json;
-use serde_wasm_bindgen;
+use serde_wasm_bindgen::{self};
```
bnjbvr, line 25: nit: this is redundant, as it's equivalent to the previous form. Doesn't Clippy complain about this?

src/machine.rs
```diff
@@ -883,15 +885,18 @@ impl OlmMachine {
     /// `predicate` is a closure that will be called for every known
     /// `InboundGroupSession`, which represents a room key. If the closure
     /// returns `true`, the `InboundGroupSession` will be included in the
-    /// export, otherwise it won't.
+    /// export; otherwise it won't.
+    ///
+    /// Returns a Promise containing a Result containing a String which is a
+    /// JSON-encoded array of ExportedRoomKey objects.
```
bnjbvr, line 891: (Unrelated to this specific PR, but) I wonder if we could make some kind of newtype that has phantom data of the right type, and when used with `wasm_bindgen` would be equivalent to a `Promise`. That way, we could have tighter checks for the return values of these functions, and we wouldn't need comments like this one, which has a high risk of getting outdated as soon as it lands.

src/machine.rs
```diff
@@ -1572,3 +1578,19 @@ pub(crate) async fn promise_result_to_future(
...
+async fn stream_to_json_array<T, S>(mut stream: Pin<&mut S>) -> Result<String, anyhow::Error>
+where
+    T: Serialize,
+    S: Stream<Item = T>,
+{
+    let mut stream_json = vec![];
+    let mut ser = serde_json::Serializer::new(&mut stream_json);
+    let mut seq = ser.serialize_seq(None).map_err(|e| anyhow::Error::msg(e.to_string()))?;
+    while let Some(key) = stream.next().await {
+        seq.serialize_element(&key).map_err(|e| anyhow::Error::msg(e.to_string()))?;
```
bnjbvr, line 1591: nit: for both, can `anyhow` not figure out how to wrap the error?

```suggestion
    let mut seq = ser.serialize_seq(None)?;
    while let Some(key) = stream.next().await {
        seq.serialize_element(&key)?;
```

src/machine.rs
```diff
@@ -1572,3 +1578,19 @@ pub(crate) async fn promise_result_to_future(
...
+    }
+    seq.end().map_err(|e| anyhow::Error::msg(e.to_string()))?;
```
bnjbvr, line 1593: (here too actually)

```suggestion
    seq.end()?;
```

src/machine.rs
```diff
@@ -1572,3 +1578,19 @@ pub(crate) async fn promise_result_to_future(
         }
     }
 }
+
+async fn stream_to_json_array<T, S>(mut stream: Pin<&mut S>) -> Result<String, anyhow::Error>
+where
+    T: Serialize,
+    S: Stream<Item = T>,
+{
+    let mut stream_json = vec![];
```
bnjbvr, line 1587: Would initializing with an initial capacity make sense? Assuming the stream's not empty, we can likely avoid many reallocations by having a large initial size here?

src/machine.rs
bnjbvr: Also: is it a breaking change? I don't think so, because the previous code seemed to return a similar type, but maybe it did change :eyes: If it did, is it worth adding a note to the changelog?

bnjbvr (approved):
lgtm. Is it already tested or could it benefit from a new test, at least validating what the return type actually is? Thanks!

src/machine.rs
andybalaam: Weirdly, Clippy doesn't complain, but I will fix it. No idea how it appeared.

src/machine.rs
andybalaam: The return type is unchanged, just better documented :-)

src/machine.rs
andybalaam: I think so, but I don't know how to estimate a good size here. Do we do anything similar anywhere else in the code? Do we have any rules of thumb?

andybalaam:
> lgtm. Is it already tested or could it benefit from a new test, at least validating what the return type actually is? Thanks!

It's already tested in [machine.test.ts](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/blob/main/tests/machine.test.ts#L734).

src/machine.rs
andybalaam: Brilliant, thank you - done

PR #117, opened 2024-05-15
Add a constructor for the Curve25519Public key type

poljar:
I took the opportunity to rename the vodozemac module as well, since it will soon get some more functionality for the QR login support.

PR #116, opened 2024-05-15
Bump the version of the SDK and a minor fix to silence a deprecation warning

poljar:
This bumps the version of the SDK to include the required pieces for the QR login support. Exposing the QR login structs and methods will come in a separate PR.

PR #118, opened 2024-05-15
Expose QR login related structs and methods

poljar:
This PR depends on #116  and #117 so opening it as a draft. It adds all the functionality necessary to make QR code login work in the js-sdk.

A review commit by commit is recommended.

### DIFF
--- a/src/qr_login.rs
+++ b/src/qr_login.rs
@@ -1,0 +1,141 @@
+//! Types for QR code login
+
+use matrix_sdk_crypto::types::qr_login;
+use url::Url;
+use wasm_bindgen::prelude::*;
+
+use crate::vodozemac::Curve25519PublicKey;
+
+/// The mode of the QR code login.
+///
+/// The QR code login mechanism supports both, the new device, as well as the
+/// existing device to display the QR code.
+///
+/// The different modes have an explicit one-byte identifier which gets added to
+/// the QR code data.
+#[wasm_bindgen]
+#[derive(Debug)]
+pub enum QrCodeMode {
+    /// The new device is displaying the QR code.
+    Login,
+    /// The existing device is displaying the QR code.
+    Reciprocate,
+}
+
+impl From<qr_login::QrCodeMode> for QrCodeMode {
+    fn from(value: qr_login::QrCodeMode) -> Self {
+        match value {
+            qr_login::QrCodeMode::Login => Self::Login,
+            qr_login::QrCodeMode::Reciprocate => Self::Reciprocate,
+        }
+    }
+}
+
+/// Data for the QR code login mechanism.
+///
+/// The [`QrCodeData`] can be serialized and encoded as a QR code or it can be
+/// decoded from a QR code.
+#[wasm_bindgen]
+#[derive(Debug)]
+pub struct QrCodeData {
+    inner: qr_login::QrCodeData,
+}
+
+#[wasm_bindgen]
+impl QrCodeData {
+    /// Create new [`QrCodeData`] from a given public key, a rendezvous URL and,
+    /// optionally, a homeserver URL.
+    ///
+    /// If a homeserver URL is given, then the [`QrCodeData`] mode will be
+    /// [`QrCodeMode::Reciprocate`], i.e. the QR code will contain data for the
+    /// existing device to display the QR code.
+    ///
+    /// If no homeserver is given, the [`QrCodeData`] mode will be
+    /// [`QrCodeMode::Login`], i.e. the QR code will contain data for the
+    /// new device to display the QR code.
+    #[wasm_bindgen(constructor)]
+    pub fn new(
+        public_key: Curve25519PublicKey,
+        rendezvous_url: &str,
+        homeserver_url: Option<String>,
+    ) -> Result<QrCodeData, JsError> {
+        let public_key = public_key.inner;
+        let rendezvous_url = Url::parse(rendezvous_url)?;
+
+        let mode_data = if let Some(homeserver_url) = homeserver_url {
+            qr_login::QrCodeModeData::Reciprocate { homeserver_url: Url::parse(&homeserver_url)? }
+        } else {
+            qr_login::QrCodeModeData::Login
+        };
+
+        let inner = qr_login::QrCodeData { public_key, rendezvous_url, mode_data };
+
+        Ok(QrCodeData { inner })
+    }
+
+    /// Attempt to decode a slice of bytes into a [`QrCodeData`] object.
+    ///
+    /// The slice of bytes would generally be returned by a QR code decoder.
+    pub fn from_bytes(bytes: &[u8]) -> Result<QrCodeData, JsError> {
+        Ok(Self { inner: qr_login::QrCodeData::from_bytes(bytes)? })
+    }
+
+    /// Encode the [`QrCodeData`] into a list of bytes.
+    ///
+    /// The list of bytes can be used by a QR code generator to create an image
+    /// containing a QR code.
+    pub fn to_bytes(&self) -> Vec<u8> {
+        self.inner.to_bytes()
+    }
+
+    /// Attempt to decode a base64 encoded string into a [`QrCodeData`] object.
+    pub fn from_base64(data: &str) -> Result<QrCodeData, JsError> {
+        Ok(Self { inner: qr_login::QrCodeData::from_base64(data)? })
+    }
+
+    /// Encode the [`QrCodeData`] into a string using base64.
+    ///
+    /// This format can be used for debugging purposes and the
+    /// [`QrcodeData::from_base64()`] method can be used to parse the string
+    /// again.
+    pub fn to_base64(&self) -> String {
+        self.inner.to_base64()
+    }
+
+    /// Get the Curve25519 public key embedded in the [`QrCodeData`].
+    ///
+    /// This Curve25519 public key should be used to establish an
+    /// [ECIES](https://en.wikipedia.org/wiki/Integrated_Encryption_Scheme)
+    /// (Elliptic Curve Integrated Encryption Scheme) channel with the other
+    /// device.
+    #[wasm_bindgen(getter)]
+    pub fn public_key(&self) -> Curve25519PublicKey {
+        self.inner.public_key.into()
+    }
+
+    /// Get the URL of the rendezvous server which will be used to exchange
+    /// messages between the two devices.
+    #[wasm_bindgen(getter)]
+    pub fn rendezvous_url(&self) -> String {
+        self.inner.rendezvous_url.as_str().to_owned()
+    }
+
+    /// Get the homeserver URL which the new device will be logged in to.
+    ///
+    /// This will be only available if the existing device has generated the QR
+    /// code and the new device is the one scanning the QR code.
+    #[wasm_bindgen(getter)]
+    pub fn homeserver_url(&self) -> Option<String> {
+        if let qr_login::QrCodeModeData::Reciprocate { homeserver_url } = &self.inner.mode_data {
+            Some(homeserver_url.as_str().to_owned())
+        } else {
+            None
+        }
+    }
+
+    /// Get the mode of this [`QrCodeData`] instance.
+    #[wasm_bindgen(getter)]
+    pub fn mode(&self) -> QrCodeMode {
+        self.inner.mode().into()
+    }
+}
--- a/src/store.rs
+++ b/src/store.rs
@@ -2,7 +2,10 @@
 
 use std::sync::Arc;
 
-use matrix_sdk_crypto::store::{DynCryptoStore, IntoCryptoStore, MemoryStore};
+use matrix_sdk_crypto::{
+    store::{DynCryptoStore, IntoCryptoStore, MemoryStore},
+    types::BackupSecrets,
+};
 use wasm_bindgen::prelude::*;
 
 use crate::{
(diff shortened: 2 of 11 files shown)


src/qr_login.rs
```diff
@@ -0,0 +1,140 @@
...
+/// the QR code data.
+#[wasm_bindgen]
+#[derive(Debug)]
+pub enum QrCodeMode {
+    /// Enum variant for the case where the new device is displaying the QR
+    /// code.
+    Login,
+    /// Enum variant for the case where the existing device is displaying the QR
+    /// code.
+    Reciprocate,
```
bnjbvr, line 24: I find these names a bit confusing, but maybe they're referring to names used in the spec? If not, could we consider giving them some names that are more eloquent, like `QrCodeMode` => `QrCodeLoginOrigin` (or s/origin/source), `Login` => `NewDevice`, `Reciprocate` => `ExistingDevice`?

Independently of the above question: it's weird to me to have `Enum variant for the case where` in the doc comment, since this applies to all the enum variants in the world.  Could we get rid of that please?

src/qr_login.rs
```diff
@@ -0,0 +1,140 @@
...
+#[wasm_bindgen]
+#[derive(Debug)]
+pub struct QrCodeData {
+    inner: qr_login::QrCodeData,
+}
+
+#[wasm_bindgen]
+impl QrCodeData {
+    /// Create new [`QrCodeData`] from a given public key, a rendezvous URL and,
+    /// optionally, a homeserver ULR.
```
bnjbvr, line 49: ULR => URL

src/qr_login.rs
```diff
@@ -0,0 +1,140 @@
...
+        Ok(Self { inner: qr_login::QrCodeData::from_base64(data)? })
+    }
+
+    /// Encode the [`QrCodeData`] into a list of bytes.
+    ///
+    /// The list of bytes can be used by a QR code generator to create an image
+    /// containing a QR code.
+    pub fn to_base64(&self) -> String {
+        self.inner.to_base64()
+    }
```
bnjbvr, line 104: This is the same doc comment as `to_bytes()`. Can you update the doc comment and explain how it's different from the other one? Or maybe one of the two methods can be removed?

src/qr_login.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    /// Get the Curve25519 public key embedded in the [`QrCodeData`].
+    ///
+    /// This Curve25519 public key should be used to establish an ECIES channel
```
bnjbvr, line 108: Can you explicit the ECIES acronym once, or point to a doc / link that does?

src/qr_login.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// with the other device.
+    #[wasm_bindgen(getter)]
+    pub fn public_key(&self) -> Curve25519PublicKey {
+        self.inner.public_key.into()
+    }
+
+    /// Get the URL of the rendezvous server which will be used to exchange
+    /// devices between the two devices.
```
bnjbvr, line 116: exchange devices between the two devices? :raised_eyebrow:

src/vodozemac/ecies.rs
```diff
@@ -0,0 +1,209 @@
...
+pub struct Ecies {
+    inner: Option<ecies::Ecies>,
+    public_key: Curve25519PublicKey,
+}
+
+#[wasm_bindgen]
+impl Ecies {
+    /// Create a new, random, unestablished ECIES session.
+    ///
+    /// This method will use the `MATRIX_QR_CODE_LOGIN` info.
```
bnjbvr, line 44: What is this constant referring to? Can we use an intra doc link to make it more explicit?

src/vodozemac/ecies.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// other side.
+    pub fn establish_inbound_channel(
+        &mut self,
+        initial_message: &str,
+    ) -> Result<InboundCreationResult, JsError> {
+        let message = ecies::InitialMessage::decode(&initial_message)?;
+        let result = self
+            .inner
+            .take()
+            .ok_or_else(Self::used_up_error)?
```
bnjbvr, line 75: Can this method take `mut self` instead of `&mut self` so that it's impossible to hit the `used_up_error` and we can get rid of it?

src/vodozemac/ecies.rs
```diff
@@ -0,0 +1,209 @@
...
+            .establish_inbound_channel(&message)?;
```
bnjbvr, line 76: If the answer to the above question is "no", then: is it worth putting the `self.inner.take()` back into the `self.inner`, if this line yields an error result?

src/vodozemac/ecies.rs
```diff
@@ -0,0 +1,209 @@
...
+    pub fn establish_outbound_channel(
+        &mut self,
+        public_key: &Curve25519PublicKey,
+        initial_message: &str,
+    ) -> Result<OutboundCreationResult, JsError> {
+        let result = self
+            .inner
+            .take()
+            .ok_or_else(Self::used_up_error)?
+            .establish_outbound_channel(public_key.inner, initial_message.as_bytes())?;
```
bnjbvr, line 98: Same two remarks.

src/vodozemac/ecies.rs
```diff
@@ -0,0 +1,209 @@
...
+}
+
+/// An established ECIES session.
+///
+/// This session can be used to encrypt and decrypt messages between the two
+/// sides of the channel.
+#[derive(Clone)]
+#[wasm_bindgen]
+pub struct EstablishedEcies {
+    inner: Arc<Mutex<ecies::EstablishedEcies>>,
```
bnjbvr, line 114: Out of curiosity, are the arc/mutex needed here, since this is wasm and in theory mono-threaded? (unless the consumer is making use of shared array buffers)

src/vodozemac/ecies.rs
```diff
@@ -0,0 +1,209 @@
...
+        let message = ecies::Message::decode(message)?;
+        let result = self.inner.lock().unwrap().decrypt(&message)?;
+
+        Ok(String::from_utf8_lossy(&result).to_string())
+    }
+
+    /// Get the [`CheckCode`] which uniquely identifies this
+    /// [`EstablishedEcies`] session.
+    ///
+    /// This check code can be used to check that both sides of the session are
```
bnjbvr, line 143: nit: "this check code can be used to check" reads weird, can you tweak the wording please?

src/vodozemac/ecies.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// indeed using the same shared secret.
+    pub fn check_code(&self) -> CheckCode {
+        self.inner.lock().unwrap().check_code().into()
+    }
+}
+
+/// A check code that can be used to confirm that two [`EstablishedEcies`]
+/// objects share the same secret. This is supposed to be shared out-of-band to
+/// protect against active MITM attacks.
```
bnjbvr, line 152: nit: please explicit MITM

src/store.rs
```diff
@@ -170,3 +173,69 @@ impl RoomKeyInfo {
         self.inner.session_id.clone()
     }
 }
+
+/// Struct containing the bundle of secrets to fully activate a new devices for
```
bnjbvr, line 177: nit: new devices => new device

tests/ecies.test.ts
```diff
@@ -26,3 +26,65 @@ describe(Ecies.name, () => {
...
+        const secondDevice = new DeviceId("DEVICE2");
+
+        const firstMachine = await OlmMachine.initialize(userId, firstDevice);
+        const secondMachine = await OlmMachine.initialize(userId, secondDevice);
+
+        await firstMachine.bootstrapCrossSigning(false);
+        const bundle = await firstMachine.exportSecretsBundle();
+
+        const alice = new Ecies();
+        const bob = new Ecies();
```
bnjbvr, line 64: is it really a bob or rather an alice2?

bnjbvr (approved):
Looks good to me. My comments are mostly nits or suggestions for renaming, so nothing that should prevent from merging this PR once those comments are addressed. Thanks!

tests/ecies.test.ts
poljar: It's Alice's second device called Bob.

src/store.rs
```diff
@@ -170,3 +173,69 @@ impl RoomKeyInfo {
...
+#[wasm_bindgen]
+pub struct SecretsBundle {
+    pub(super) inner: matrix_sdk_crypto::types::SecretsBundle,
+}
+
+/// The backup-specific parts of a secrets bundle.
+#[derive(Debug)]
+#[wasm_bindgen(getter_with_clone)]
+pub struct BackupSecretsBundle {
+    /// The backup decryption key.
```
BillCarsonFr, line 189: Doc: What's the format? base64 or prefix + base58 (recovery key export format)?

BillCarsonFr (approved):
Just a comment on doc

src/qr_login.rs
poljar: The terms are from the MSC: https://github.com/matrix-org/matrix-spec-proposals/blob/element-hq/oidc-qr-login/proposals/4108-oidc-qr-login.md#qr-code-format. I did shorten them from "initiate login" and "reciprocate login" to "Login" and "Reciprocate" though. I can use the longer names here but I would stick to the names used in the spec.

The docs on the enum variants were shortened in e922f0813877d4841c138585bd2cb01fa462437f.

src/qr_login.rs
poljar: We do link to the wiki page for ECIES in the vodozemac/ecies module, and we deffer to the vodozemac docs to not copy paste everything from vodozemac. With that in mind do you want me to link to the wiki page again here?

bnjbvr: Ah thanks, maybe we could explicit the acronym once at least?

src/vodozemac/ecies.rs
poljar: The constant is referring to the constant used in vodozemac and defined in the MSC. We can't use an intra-doc link here. I'll expand the comment a bit, it makes less sense compared to the vodozemac one.

src/vodozemac/ecies.rs
poljar: No, that doesn't work with WASM.

src/vodozemac/ecies.rs
poljar: No, the channel is supposed to be ephemeral for security reasons, and if things fail you shouldn't try to reuse the generated keys.

src/vodozemac/ecies.rs
poljar: AFAIR yes, leaving them out did not work. On WASM they will be compiled into a no-op, but the Rust compiler complains about... Err I forgot when/where it complained.

src/qr_login.rs
poljar: 4be534b3035a8bb96192fdfe6b9c2e656f0d3bef

src/qr_login.rs
poljar: 4be534b3035a8bb96192fdfe6b9c2e656f0d3bef

src/qr_login.rs
poljar: Added the link and defined the acronym: 4be534b3035a8bb96192fdfe6b9c2e656f0d3bef.

src/qr_login.rs
poljar: Yup, a brain fart: 4be534b3035a8bb96192fdfe6b9c2e656f0d3bef.

src/store.rs
poljar: Fixed in: 94020b69dde69edf6963b8ad60d3af1b10ed18b1.

src/store.rs
poljar: It's unpadded base64, just like in the `m.secret.send` event. Added it to the docs in: 94020b69dde69edf6963b8ad60d3af1b10ed18b1.

src/vodozemac/ecies.rs
poljar: Added here: 86f43b7f619aaf0d42e252bdb270db8bba22bd9d.

src/vodozemac/ecies.rs
poljar: Same answer as above.

src/vodozemac/ecies.rs
poljar: Reworded in 86f43b7f619aaf0d42e252bdb270db8bba22bd9d.

src/vodozemac/ecies.rs
poljar: Did so in: 86f43b7f619aaf0d42e252bdb270db8bba22bd9d.

src/qr_login.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// Create new [`QrCodeData`] from a given public key, a rendezvous URL and,
+    /// optionally, a homeserver URL.
+    ///
+    /// If a homeserver URL is given, then the [`QrCodeData`] mode will be
+    /// [`QrCodeMode::Reciprocate`], i.e. the QR code will contain data for the
+    /// existing device to display the QR code.
+    ///
+    /// If no homeserver is given, the [`QrCodeData`] mode will be
+    /// [`QrCodeMode::Login`], i.e. the QR code will contain data for the
+    /// new device to display the QR code.
```
richvdh, line 55: please bear in mind that `` [`QrCodeData`] `` doesn't work in the documentation generated by `wasm_bindgen`: see https://matrix-org.github.io/matrix-rust-sdk-crypto-wasm/classes/QrCodeData.html#constructor.

You need to use tsdoc-style `{@link ... }` links.

src/qr_login.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// This will be only available if the existing device has generated the QR
+    /// code and the new device is the one scanning the QR code.
+    #[wasm_bindgen(getter)]
+    pub fn homeserver_url(&self) -> Option<String> {
+        if let qr_login::QrCodeModeData::Reciprocate { homeserver_url } = &self.inner.mode_data {
+            Some(homeserver_url.as_str().to_owned())
+        } else {
+            None
+        }
+    }
```
richvdh, line 134: When we're adding new methods to the WASM bindings, please can we use `#[wasm_bindgen(js_name)]` to give them camelCase names, to be consistent with the rest of the bindings?

src/qr_login.rs
poljar: I changed the style in: https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/125/commits/2c9bd6fa0f62ef9e39e90c08292be2a7b643971c.

PR #119, opened 2024-05-16
Expose new method `OlmMachine.openWithKey`

richvdh:
... allowing the application to pass in the key directly if they have a cryptographically-secure key.

Part of https://github.com/element-hq/element-web/issues/26821. Requires https://github.com/matrix-org/matrix-rust-sdk/pull/3423 (so will fail CI for now)

### DIFF
--- a/src/store.rs
+++ b/src/store.rs
@@ -5,6 +5,7 @@ use std::sync::Arc;
 use anyhow::Context;
 use matrix_sdk_crypto::store::{DynCryptoStore, IntoCryptoStore, MemoryStore};
 use wasm_bindgen::prelude::*;
+use zeroize::{Zeroize, Zeroizing};
 
 use crate::{
     encryption::EncryptionAlgorithm, identifiers::RoomId, impl_from_to_inner,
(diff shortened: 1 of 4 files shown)


src/store.rs
```diff
@@ -95,6 +92,41 @@ impl StoreHandle {
...
+    /// * `store_key` - The key that should be used to encrypt the store, for
+    ///   IndexedDB-based stores. Must be a 32-byte array.
+    #[wasm_bindgen(js_name = "openWithKey")]
+    pub async fn open_with_key(
+        store_name: String,
+        mut store_key: Vec<u8>,
+    ) -> Result<StoreHandle, JsError> {
+        use zeroize::Zeroize;
+
+        let mut store_key_slice: [u8; 32] = store_key
```
andybalaam, line 113: This is not a slice, so maybe should be called `store_key_array`?

andybalaam (approved):
Looks good with one minor comment.

src/store.rs
richvdh: Indeed. Fixed, thanks.

PR #121, opened 2024-05-23
Fix backed-up keys being re-backed-up

richvdh:
Bump to a version of the rust SDK that includes https://github.com/matrix-org/matrix-rust-sdk/pull/3448 and https://github.com/matrix-org/matrix-rust-sdk/pull/3456, to fix https://github.com/matrix-org/matrix-rust-sdk/issues/3447.

Also, pass the backup version into `import_backed_up_room_keys` to avoid using the deprecated method on `BackupMachine`.

### DIFF
--- a/tests/machine.test.ts
+++ b/tests/machine.test.ts
@@ -1442,27 +1442,32 @@ describe(OlmMachine.name, () => {
     test("Updating devices should call devicesUpdatedCallback", async () => {
         const userId = new UserId("@alice:example.org");
         const deviceId = new DeviceId("ABCDEF");
-        const machine = await OlmMachine.initialize(userId, deviceId);
+        const firstMachine = await OlmMachine.initialize(userId, deviceId);
 
         const callback = jest.fn().mockImplementation(() => Promise.resolve(undefined));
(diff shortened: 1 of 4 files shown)


tests/machine.test.ts
```diff
@@ -1442,27 +1442,32 @@ describe(OlmMachine.name, () => {
...
 
-        const outgoingRequests = await machine.outgoingRequests();
+        const secondDeviceId = new DeviceId("GHIJKL");
+        const secondMachine = await OlmMachine.initialize(userId, secondDeviceId);
+
+        // Fish the KeysUploadRequest out of secondMachine's outgoingRequests.
         let deviceKeys;
-        // outgoingRequests will have a KeysUploadRequest before the
-        // KeysQueryRequest, so we grab the device upload and put it in the
-        // response to the /keys/query
```
richvdh, line 1454: this test didn't work any more after the rust-sdk upgrade, due to https://github.com/matrix-org/matrix-rust-sdk/pull/3442: the response no longer triggered a device update notification. Hence, we now feed the device from a second OlmMachine in.

poljar (approved):

Issue #110, opened 2023-06-26
Return types are very weak


richvdh:
Currently, `matrix-sdk-crypto-js` defines lots of functions which return types like `Promise`, rather than using typescript's support for generics to declare what that promise will contain.

For example: [`OlmMachine.getMissingSessions`](https://matrix-org.github.io/matrix-rust-sdk/bindings/matrix-sdk-crypto-js/classes/OlmMachine.html#getMissingSessions) could be defined to return `Promise<KeysClaimRequest | undefined>`.

Failure to declare these types correctly (a) makes development against the library much harder, because application developers have to add their own type annotations; (b) increases the risk of a regression due to an unnoticed change in return value

richvdh:
FWIW with modern wasm-bindgen, we can improve on this considerably: `wasm-bindgen` now understands `async` functions and will automatically turn `async fn foo() -> Bar` into javascript-side functions that return `Promise<Bar>`, provided `Bar` has an `Into<JsValue>` impl (cf https://rustwasm.github.io/wasm-bindgen/reference/js-promises-and-rust-futures.html#return-values-of-async-fn).

 We should go through and fix the existing methods that explicitly call `future_to_promise`, and stop adding new ones

Hywan:
Oh nice!

richvdh:
> FWIW with modern wasm-bindgen, we can improve on this considerably: `wasm-bindgen` now understands `async` functions and will automatically turn `async fn foo() -> Bar` into javascript-side functions that return `Promise<Bar>`, provided `Bar` has an `Into<JsValue>` impl (cf [rustwasm.github.io/wasm-bindgen/reference/js-promises-and-rust-futures.html#return-values-of-async-fn](https://rustwasm.github.io/wasm-bindgen/reference/js-promises-and-rust-futures.html#return-values-of-async-fn)).
> 
> We should go through and fix the existing methods that explicitly call `future_to_promise`, and stop adding new ones.

It turns out that wasm-bindgen's `async` support doesn't play nicely with explicit calls to `.close`/`.free`: https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/205, so this doesn't work for us.

[...]

PR #123, opened 2024-06-03
Improve types and docs of `{import,export}SecretsBundle`

richvdh:
Followup on #118

cf #110

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -534,7 +534,8 @@ impl OlmMachine {
         })
     }
 
-    /// Export all the secrets we have in the store into a [`SecretsBundle`].
+    /// Export all the secrets we have in the store into a {@link
+    /// SecretsBundle}.
     ///
     /// This method will export all the private cross-signing keys and, if
     /// available, the private part of a backup key and its accompanying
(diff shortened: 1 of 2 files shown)


richvdh:
cc @poljar since you wrote the original PR

src/machine.rs
```diff
@@ -534,7 +534,8 @@ impl OlmMachine {
         })
     }
 
-    /// Export all the secrets we have in the store into a [`SecretsBundle`].
+    /// Export all the secrets we have in the store into a {@link
+    /// SecretsBundle}.
```
richvdh, line 538: Since this documentation is used by typedoc, we have to use link formats that it understands.

(Note that currently the documentation at https://matrix-org.github.io/matrix-rust-sdk-crypto-wasm/classes/OlmMachine.html#importSecretsBundle doesn't link correctly).

src/machine.rs
```diff
@@ -546,35 +547,29 @@ impl OlmMachine {
...
     /// device is sharing this with a new device. The imported cross-signing
-    /// keys will create a [`OwnUserIdentity`] and mark it as verified.
+    /// keys will create a {@link OwnUserIdentity} and mark it as verified.
     ///
     /// The backup key will be persisted in the store and can be enabled using
-    /// the [`BackupMachine`].
+    /// the {@link BackupMachine}.
+    ///
+    /// The provided `SecretsBundle` is freed by this method; be careful not to
+    /// use it once this method has been called.
```
richvdh, line 568: invalidating an argument is a very non-JS thing to do, so let's call it out explicitly.

BillCarsonFr (approved):
Very nice improvement!

PR #124, opened 2024-06-13
Update the matrix-sdk crates to 9b05d0d822

poljar:


Hywan (approved):
LGTM, thanks!

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # UNRELEASED
 
+**BREAKING CHANGES**
+
+-   Update matrix-rust-sdk to `9b05d0d82`, which includes:
+
+    -   `Device.requestVerification`, `UserIdentities.requestVerification`, and
+        `UserIdentities.verificationRequestContent` is not async anymore.
+        ([#3513](https://github.com/matrix-org/matrix-rust-sdk/pull/3513))
```
richvdh, line 9: Worth mentioning that converting an async function to not-async is not normally a breaking change in JS (though typescript might complain at you about the redundant `await`).

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+
+    -   Use the server name in the `QrCodeData` instead of the homeserver URL.
+        ([#3537](https://github.com/matrix-org/matrix-rust-sdk/pull/3537))
```
richvdh, line 12: Not sure if this is a breaking change either?

CHANGELOG.md
poljar: Oh that's unexpected, thanks for letting me know.

CHANGELOG.md
poljar: Well we did rename the field, and the type of the field, thus the getter in the bindings was also renamed.

CHANGELOG.md
t3chguy: > Worth mentioning that converting an async function to not-async is not normally a breaking change in JS (though typescript might complain at you about the redundant await).

I disagree, as the method return value loses `.then` and `.catch` properties for non-async-await envs.

CHANGELOG.md
richvdh: ah, it might be good to call out that rename then?

poljar: https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/125

PR #134, opened 2024-07-04
Update Rust SDK to latest version

uhoreg:
Looks like three different matrix-rust-sdk PRs introduced incompatible changes.  This is my attempt to fix them.  I'm certain that the fix for my PR is correct, but fairly sure that the `sender_data` fix is correct, and less sure that the `CollectStrategy` fix is sensible.

This introduces a change that will break js-sdk (yay for cascading breakage), so that will need to be updated.

~Not sure why the test is failing.  Seems to be unrelated to the changes here, so must be some change elsewhere.  I'm guessing that it should be fine to update the expected values to the actual values, but someone who knows about why the discrepancy is there should verify.~


richvdh:
Fixes fixes https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/134

src/encryption.rs
```diff
@@ -116,6 +116,51 @@ impl From<matrix_sdk_crypto::types::EventEncryptionAlgorithm> for EncryptionAlgo
     }
 }
 
+#[wasm_bindgen()]
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// Strategy to collect the devices that should receive room keys for the
+/// current discussion.
+///
+/// See matrix_sdk_crypto::CollectStrategy
```
richvdh, line 124: Please make these docs stand in their own right, so that the online doc (https://matrix-org.github.io/matrix-rust-sdk-crypto-wasm/classes/OlmMachine.html) is useful. (The fact that there are existing docs that just point to a Rust class is definitely bad.)

src/encryption.rs
```diff
@@ -33,8 +33,8 @@ pub struct EncryptionSettings {
 
     /// Should untrusted devices receive the room key, or should they be
     /// excluded from the conversation.
-    #[wasm_bindgen(js_name = "onlyAllowTrustedDevices")]
-    pub only_allow_trusted_devices: bool,
+    #[wasm_bindgen(js_name = "sharingStrategy")]
+    pub sharing_strategy: CollectStrategy,
```
richvdh, line 37: It would be good to update the tests to test reading, and modifying, this.

richvdh (changes requested):
Could you update the changelog pls?

src/encryption.rs
```diff
@@ -116,6 +116,51 @@ impl From<matrix_sdk_crypto::types::EventEncryptionAlgorithm> for EncryptionAlgo
     }
 }
 
+#[wasm_bindgen()]
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// Strategy to collect the devices that should receive room keys for the
```
Hywan, line 121: Please put the `#[…]` after the documentation. It's more Rust idiomatic to do so.

src/encryption.rs
uhoreg: Added test

src/encryption.rs
uhoreg: done

src/encryption.rs
uhoreg: done

uhoreg:
> Could you update the changelog pls?

I've added a changelog.  I'm not sure what the criteria are for which rust-sdk-crypto changes to highlight in the changelogs, so I just highlighted the one that causes the breaking change.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
 
+**BREAKING CHANGES**
+
+-   `EncryptionSettings.onlyAllowTrustedDevices` has been replaced with
+    `EncryptionSettings.sharingStrategy`, which adds the ability to share only
+    with cross-signed devices.
+
+-   Update matrix-rust-sdk to `11cbf849c`, which includes:
+
+    -   refactor(sdk-crypto): Room key sharing, introduce extensible strategy ([#3605](https://github.com/matrix-org/matrix-rust-sdk/pull/3605))
```
richvdh, line 11: Ideally we'd include everything here which affects the behaviour of the library. That should just be a matter of diffing the changelogs in the rust-sdk, but people are terrible at remembering to update them.

```suggestion
**Other changes**

-   Update matrix-rust-sdk to `11cbf849c`, which includes:

    -   refactor(sdk-crypto): Room key sharing, introduce extensible strategy
        ([#3605](https://github.com/matrix-org/matrix-rust-sdk/pull/3605))
    
    -   Log the content of received `m.room_key.withheld` to-device events.
        ([#3591](https://github.com/matrix-org/matrix-rust-sdk/pull/3591))

    -   Attempt to decrypt bundled events (reactions and the latest thread reply) if they are found in the unsigned part of an event.
        ([#3468](https://github.com/matrix-org/matrix-rust-sdk/pull/3468))

```

richvdh (approved):

PR #138, opened 2024-07-25
add DCO

joshsimmons:
The Matrix.org Foundation is rolling out this DCO across all public repositories in its namespace. Of note, this DCO does not require the use of "real" or "legally identifiable" names.


andybalaam (approved):
Thanks for making this very easy for us!

PR #143, opened 2024-09-13
Add mark_all_tracked_users_as_dirty

kegsay:
This is required for SSS in Element-Web. See https://github.com/matrix-org/matrix-rust-sdk/pull/3965 for more info.

Had to run `cargo update matrix-sdk-crypto` to pull in the latest code with this function. Hopefully I've done it right.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
     clients as the server may give up trying to persist device list updates for
     the client at some point, after which the client must treat all devices as dirty.
 
+-   Update matrix-rust-sdk to `2408df8bf`. No changes relevant to these bindings.
+
 # matrix-sdk-crypto-wasm v8.0.0
 
 **BREAKING CHANGES**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # UNRELEASED
 
+**Other changes**
+
+-   Add `OlmMachine.markAllTrackedUsersAsDirty` to invalidate the device lists
+    for all known users. This is required for [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186)
+    clients as the server may give up trying to persist device list updates for
+    the client at some point, after which the client must treat all devices as dirty.
+
 # matrix-sdk-crypto-wasm v8.0.0
 
 **BREAKING CHANGES**
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -178,6 +178,20 @@ impl OlmMachine {
         future_to_promise(async move { Ok(me.display_name().await?) })
     }
 
+    /// Mark all tracked users as dirty.
+    ///
+    /// All users *whose device lists we are tracking* are flagged as needing a
+    /// key query. Users whose devices we are not tracking are ignored.
+    #[wasm_bindgen(js_name = "markAllTrackedUsersAsDirty")]
+    pub fn mark_all_tracked_users_as_dirty(&self) -> Promise {
+        let machine = self.inner.clone();
+
+        future_to_promise(async move {
+            machine.mark_all_tracked_users_as_dirty().await?;
+            Ok(JsValue::UNDEFINED)
+        })
+    }
+
     /// Whether automatic transmission of room key requests is enabled.
     ///
     /// Room key requests allow the device to request room keys that it might
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -178,20 +178,6 @@ impl OlmMachine {
         future_to_promise(async move { Ok(me.display_name().await?) })
     }
 
-    /// Mark all tracked users as dirty.
-    ///
-    /// All users *whose device lists we are tracking* are flagged as needing a
-    /// key query. Users whose devices we are not tracking are ignored.
-    #[wasm_bindgen(js_name = "markAllTrackedUsersAsDirty")]
-    pub fn mark_all_tracked_users_as_dirty(&self) -> Promise {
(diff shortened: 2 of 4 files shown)


richvdh:
Could you update the changelog, please

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # UNRELEASED
 
+**Other changes**
+
+-   Add `OlmMachine.markAllTrackedUsersAsDirty` to invalidate the device lists
+    for all known users. This is required for [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186)
+    clients as the server may give up trying to persist device list updates for
+    the client at some point, after which the client must treat all devices as dirty.
```
richvdh, line 8: could you also document the matrix-rust-sdk update, including any changes that the update may have caused and may be relevant to the application. See below for examples of the convention.

(And yes, this is a massive faff. Please join my campaign for proper, documented, releases of matrix-rust-sdk!)

src/machine.rs
```diff
@@ -178,6 +178,20 @@ impl OlmMachine {
...
+    /// key query. Users whose devices we are not tracking are ignored.
+    #[wasm_bindgen(js_name = "markAllTrackedUsersAsDirty")]
+    pub fn mark_all_tracked_users_as_dirty(&self) -> Promise {
+        let machine = self.inner.clone();
+
+        future_to_promise(async move {
+            machine.mark_all_tracked_users_as_dirty().await?;
+            Ok(JsValue::UNDEFINED)
+        })
+    }
```
richvdh, line 193: Turns out we don't need to call future_to_promise explicitly nowadays: we can just define an `async` fn and wasm-bindgen will do the work of turning the Rust Futures into JS Promises. We end up with cleaner code and better types in the generated TS.

I think this will work:

```suggestion
    pub async fn mark_all_tracked_users_as_dirty(&self) -> Result<(), JsError> {
        self.inner.mark_all_tracked_users_as_dirty().await?;
        Ok(())
    }
```

src/machine.rs
```diff
@@ -178,6 +178,20 @@ impl OlmMachine {
         future_to_promise(async move { Ok(me.display_name().await?) })
     }
 
+    /// Mark all tracked users as dirty.
+    ///
+    /// All users *whose device lists we are tracking* are flagged as needing a
+    /// key query. Users whose devices we are not tracking are ignored.
+    #[wasm_bindgen(js_name = "markAllTrackedUsersAsDirty")]
+    pub fn mark_all_tracked_users_as_dirty(&self) -> Promise {
```
richvdh, line 186: by way of making sure that this gives an ergonomic interface on the JS side, could you add something to `tests/machine.test.ts` that calls this? Doesn't have to do anything fancy, I'd just like us to call the method.

richvdh, line 186: did you give any particular thought to where in the file to add this method? It looks a bit random at first glance, but I've not taken the time to figure out if there is anywhere better.

richvdh (changes requested):

CHANGELOG.md
kegsay: Uh, how should I do this? Literally look at the new commits on rust SDK?

src/machine.rs
richvdh: somewhere near the other functions dealing with tracked users seems sensible?

CHANGELOG.md
richvdh: Pretty much. You might want to only look at commits that affect the crates we're using. And `matrix-sdk-crypto` and `matrix-sdk-indexeddb` have changelogs, so you could just look at the changes to their changelogs.

I use this very hacky script:

```
#!/bin/bash
#
# Usage: diff-for-wasm.sh <commit> [<commit>]

changelog_crates=(
    matrix-sdk-indexeddb
    matrix-sdk-crypto
)

other_crates=(
    matrix-sdk-common
    matrix-sdk-store-encryption
    matrix-sdk-qrcode
)

args=()
for crate in "${changelog_crates[@]}"; do
    args+=("./crates/${crate}/CHANGELOG.md")
done
for crate in "${other_crates[@]}"; do
    args+=("./crates/${crate}")
done
exec git diff "$@" "${args[@]}"
```

bnjbvr:
(removing my review request since @richvdh seems to be on it, thanks!)

CHANGELOG.md
kegsay: Unsure how to use your script, but looking at https://github.com/matrix-org/matrix-rust-sdk/compare/07aa6d7bc...2408df8b there isn't anything worth mentioning afaict?

kegsay: Maybe something about https://github.com/matrix-org/matrix-rust-sdk/pull/3986 ? It doesn't seem to change the API though?

richvdh: Yeah agreed, there's nothing terribly important here. Thanks for checking. Could you just add a one-liner then:

```suggestion
-   Add `OlmMachine.markAllTrackedUsersAsDirty` to invalidate the device lists
    for all known users. This is required for [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186)
    clients as the server may give up trying to persist device list updates for
    the client at some point, after which the client must treat all devices as dirty.

-   Update matrix-rust-sdk to `2408df8bf`. No changes relevant to these bindings.
```

richvdh: in fact, I'll do that now

richvdh (approved):

PR #148, opened 2024-09-30
Prepare release v9.1.0

richvdh:


bnjbvr (approved):
Thanks! (no checked-in lock files that also gets the updated version number?)

richvdh:
only the changelog, which I already did (slightly prematurely)

PR #149, opened 2024-10-04
Bring up-to-date with latest matrix-sdk (5d46b35d)

andybalaam:
In my quest to make method names consistent I first needed to reflect previous changes from matrix-sdk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # UNRELEASED
 
+**BREAKING CHANGES**
+
+-   Rename `DecryptionErrorCode.SenderIdentityPreviouslyVerified` to
+    `SenderIdentityVerificationViolation` (in line with changes to
+    matrix-rust-sdk).
+
+-   Rename `UserIdentity` to `OtherUserIdentity` (in line with changes
+    to matrix-rust-sdk).
+
+-   Update matrix-rust-sdk to `#5d46b35d`.
+
 # matrix-sdk-crypto-wasm v9.1.0
 
 -   Update matrix-rust-sdk to `866b6e5f`, which includes:
--- a/src/error.rs
+++ b/src/error.rs
@@ -22,7 +22,7 @@ pub enum DecryptionErrorCode {
     /// The sender device is not cross-signed.
     UnsignedSenderDevice,
     /// The sender's identity is unverified, but was previously verified.
-    SenderIdentityPreviouslyVerified,
+    SenderIdentityVerificationViolation,
     /// Other failure.
     UnableToDecrypt,
 }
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # UNRELEASED
 
+-   Update matrix-rust-sdk to `#5d46b35d`, which includes:
+
+    -   Rename `VerificationLevel::PreviouslyVerified` to `VerificationViolation`
+        ([#4067](https://github.com/matrix-org/matrix-rust-sdk/pull/4067)
+
+    -   Rename `UserIdentity` to `OtherUserIdentity` and
+        Rename `UserIdentities` to `UserIdentity`
+        ([#4036](https://github.com/matrix-org/matrix-rust-sdk/pull/4036))
```
richvdh, line 10: If these are changes that affect the js-side API (I think they are?), can we instead express them in terms of breaking changes to the JS objects (which don't have `::` in their names, among other things)?

richvdh (changes requested):

CHANGELOG.md
andybalaam: Thanks, remove references to non-exported things in 220d6a2

src/error.rs
```diff
@@ -22,7 +22,7 @@ pub enum DecryptionErrorCode {
     /// The sender device is not cross-signed.
     UnsignedSenderDevice,
     /// The sender's identity is unverified, but was previously verified.
-    SenderIdentityPreviouslyVerified,
+    SenderIdentityVerificationViolation,
```
richvdh, line 25: this looks like a change we need to mention in the changelog

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # UNRELEASED
 
+-   Update matrix-rust-sdk to `#5d46b35d`, which includes:
+
+    -   Rename `UserIdentity` to `OtherUserIdentity` and
+        ([#4036](https://github.com/matrix-org/matrix-rust-sdk/pull/4036))
```
richvdh, line 6: and ... ?

To be honest, this feels like it buries the lede a bit. Setting this out as a change to matrix-rust-sdk makes it feel like an internal change that a user of the library doesn't need to worry about. This is a breaking change to the *js* API; the fact it's being done to stay in sync with the rust sdk feels barely here or there. Let's set it out as a breaking change in the same way we have with previous breaking changes.

richvdh (changes requested):

src/error.rs
andybalaam: Thanks, fixed in f147076

CHANGELOG.md
andybalaam: Ah, I didn't realise this was how we were laying it out. Fixed in f147076

richvdh (approved):
There is a reference to `UserIdentity` in the documentation for `OlmMachine.get_identity` which will need an update.

LGTM otherwise.

richvdh:
(if you converted that reference to be a proper `{@link ...}`, like some of the other docs, then this sort of nonsense would be caught by CI)

PR #153, opened 2024-10-24
Create issue and ping matrix room when latest matrix-sdk code breaks the build

### LINKED ISSUE #135: Create a daily CI job that compiles against the latest rust-sdk and create an issue if broken
We keep discovering that the wasm bindings don't build against the current rust SDK. We would like a CI job which runs daily and builds and tests the wasm-bindings against the current rust SDK. If it fails, it should open an issue (if one is not already open), and ping us in a Matrix room.

https://github.com/element-hq/synapse/blob/develop/.github/workflows/twisted_trunk.yml may provide inspiration for the "open an issue" part.

andybalaam:
Please use this PR to review not just this commit, but the last few, which I pushed directly because the debugging process for github actions is awful.

You can see a proper diff in this fake PR: https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/154

It's just 2 files:

* https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/blob/andybalaam/create-issue-and-matrix-ping-on-failure/.github/latest_matrix_sdk_failed_issue_template.md and
* https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/blob/andybalaam/create-issue-and-matrix-ping-on-failure/.github/workflows/latest-matrix-sdk-crypto.yml

Thanks!

Fixes https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/135

### DIFF
--- a/.github/latest_matrix_sdk_failed_issue_template.md
+++ b/.github/latest_matrix_sdk_failed_issue_template.md
@@ -1,6 +1,11 @@
 ---
 title: Building matrix-rust-sdk-crypto-wasm against the latest matrix-sdk Rust is failing
 ---
-See https://github.com/{{env.GITHUB_REPOSITORY}}/actions/runs/{{env.GITHUB_RUN_ID}}
+Something changed in
+[matrix-rust-sdk](https://github.com/matrix-org/matrix-rust-sdk)'s crypto crate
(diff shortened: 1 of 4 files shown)


bnjbvr (approved):
lgtm, thanks for the changes!

PR #154, opened 2024-10-25
Fake PR for commenting on CI job work

andybalaam:
DO NOT MERGE

### DIFF
--- a/.github/workflows/latest-matrix-sdk-crypto.yml
+++ b/.github/workflows/latest-matrix-sdk-crypto.yml
@@ -1,0 +1,52 @@
+name: Build against latest matrix-sdk-crypto
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron:  '22 1 * * *'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  test-against-latest-matrix-sdk-crypto:
+    name: test-against-latest-matrix-sdk-crypto
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@v3
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-unknown-unknown
+
+      - name: Load cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+
+      - name: Update to the latest matrix-sdk-crypto version
+        run: cargo update matrix-sdk-crypto
+
+      - name: Install Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 20.0
+
+      - name: Install yarn
+        run: npm install --global yarn
+
+      - name: Install dependencies
+        run: yarn install
+
+      - name: Build the WebAssembly + JavaScript binding
+        run: yarn build
+
+      - name: Test the JavaScript binding
+        run: yarn test
--- a/.github/workflows/latest-matrix-sdk-crypto.yml
+++ b/.github/workflows/latest-matrix-sdk-crypto.yml
@@ -81,7 +81,7 @@ jobs:
         if: failure()
         uses: fjogeleit/http-request-action@v1
         with:
-          url: 'https://element.ems.host/hookshot/webhook/4bbf3b21-e94a-4a97-8a09-c0e03a6de173'
+          url: ${{ secrets.CRYPTO_HOOK_URL }}
           method: 'POST'
           customHeaders: '{"Content-Type": "application/json"}'
           data: '{"text": "matrix-rust-sdk-crypto-wasm is failing to build against the latest matrix-sdk Rust code. See ${{ steps.create-issue.outputs.url }}"}'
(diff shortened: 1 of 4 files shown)


.github/workflows/latest-matrix-sdk-crypto.yml
```diff
@@ -0,0 +1,87 @@
+name: Build against latest matrix-sdk-crypto
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron:  '22 1 * * *'
```
bnjbvr, line 6: Can you add a comment explaining what it means, for the peasants like me who don't speak crontab fluently? :)

.github/workflows/latest-matrix-sdk-crypto.yml
```diff
@@ -0,0 +1,87 @@
...
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+
+      - name: Update to the latest matrix-sdk-crypto version
+        run: cargo update matrix-sdk-crypto
+
+      - name: Install Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 20.0
```
bnjbvr, line 44: Does it need to be synchronized with other tasks/workflows/documentation? (If so, I try to add comments to *all* the places that need keeping in sync.)

.github/workflows/latest-matrix-sdk-crypto.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      - name: Install yarn
+        run: npm install --global yarn
```
bnjbvr, line 47: Is there a reason preferring yarn over npm? Just curious.

.github/workflows/latest-matrix-sdk-crypto.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      - name: Build the WebAssembly + JavaScript binding
+        run: yarn build
+
+      - name: Test the JavaScript binding
+        run: yarn test
+
+      - name: Open an issue on failure
+        if: failure()
+        uses: JasonEtco/create-an-issue@1b14a70e4d8dc185e5cc76d3bec9eab20257b2c5 # v2.9.2
```
bnjbvr, line 60: They haven't created a Github tag? Otherwise I'm pretty sure it's possible to use the tag name after the `@` symbol.

.github/workflows/latest-matrix-sdk-crypto.yml
```diff
@@ -0,0 +1,87 @@
...
+        if: failure()
+
+      - name: Comment on the issue
+        if: failure()
+        uses: thollander/actions-comment-pull-request@v3
+        with:
+          message: |
+            Building matrix-rust-sdk-crypto-wasm against the latest matrix-sdk Rust is failing.
+            See https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          pr-number: "${{ steps.create-issue.outputs.number }}"
```
bnjbvr, line 78: Does it mean that the first time, it will create both the issue *and* comment on it? And subsequent times, it will only comment?

(As an intuition: it might be a good thing to explain at the top of this yaml file, in a large comment, what it does overall, and how it's supposed to behave, so one doesn't have to read the entire source, maybe? With the risk that it could be desync'd from the code, indeed…)

bnjbvr (approved):
Looks good, thanks! Mostly asking for comments, or satisfying my curiosity, so I don't see anything that should block landing. Good stuff! I'm definitely going to use the auto-comment trick in some situations…

.github/workflows/latest-matrix-sdk-crypto.yml
andybalaam: The README states `yarn`, so I copied that and intend to update the other CI jobs to use yarn too, after this PR.

No idea why, except it's generally a bit nicer to use.

.github/workflows/latest-matrix-sdk-crypto.yml
andybalaam: It's best practice to use a commit hash for action version to avoid a malicious action owner being able to execute arbitrary code within our build. Ref: https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

bnjbvr: Ah, good point, TIL, thanks!

.github/workflows/latest-matrix-sdk-crypto.yml
andybalaam: f5ec776

.github/workflows/latest-matrix-sdk-crypto.yml
andybalaam: Thanks, added in f5ec776

.github/workflows/latest-matrix-sdk-crypto.yml
andybalaam: Thanks, added in f5ec776

.github/workflows/latest-matrix-sdk-crypto.yml
andybalaam: In case anyone is reading along, I realised I missed a step in the explanation - if we use a tag, the action owner (or someone who has taken over their account) can modify a tag, making it point at a different commit, whereas the commit id guarantees* we are getting the code we think we're getting.

 * except, of course, if the attacker cleverly generates a malicious commit with the same commit id, which is difficult but not impossible with SHA1.

PR #172, opened 2024-12-02
Further shrink the WASM output by using more aggressive optimizations

sandhose:
This is significantly more expensive at compilation time, but shrinks the size of the WASM file from 6MB to 5MB (2.6MB to 1.8MB compressed).


bnjbvr (approved):
Do you have any rough measurements of the compile times before/after this change? Thanks, lgtm in any case :+1:

PR #175, opened 2024-12-04
Stop deleting build results during rebuilds

richvdh:
This solves a problem, introduced in #167, with the webpack-dev-server, which would see the half-built javascript output and explode with a big error about `Can't resolve './pkg/matrix_sdk_crypto_wasm_bg.js'`

(See also #109, where a similar problem was fixed before.)


andybalaam (approved):
Looks feasible

PR #176, opened 2024-12-05
Make tsc check JS files with @ts-check annotations

sandhose:
Looks like `tsc --noEmit` wouldn't check JS files by default. This fixes that, and fixes some type errors

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,13 +20,16 @@
     ],
     "exports": {
         ".": {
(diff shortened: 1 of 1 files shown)


andybalaam (approved):
Looks good, thanks!

richvdh:
... except that CI is unhappy with you

PR #178, opened 2024-12-10
Add another entrypoint using wasm-as-esm support

richvdh:
Some environments support loading WASM modules directly as ESM, and that gives better results than the existing entry points. For example, Webpack is able to include it inside the same bundle as the rest of the sources.

### DIFF
--- a/index-wasm-esm.mjs
+++ b/index-wasm-esm.mjs
@@ -1,0 +1,39 @@
+// Copyright 2024 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// @ts-check
+
+/**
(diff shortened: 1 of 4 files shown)


index-wasm-esm.mjs
```diff
@@ -0,0 +1,39 @@
...
+ * This is the entry point for ESM environments which support the ES Module Integration Proposal for WebAssembly [1].
+ *
+ * [1]: https://github.com/webassembly/esm-integration
+ */
+
+import * as bindings from "./pkg/matrix_sdk_crypto_wasm_bg.js";
+
+/** @type {typeof import("./pkg/matrix_sdk_crypto_wasm_bg.wasm.d.ts")} */
+// @ts-expect-error TSC can't find the definitions file, for some reason.
+const wasm = await import("./pkg/matrix_sdk_crypto_wasm_bg.wasm");
```
sandhose, line 27: I'm curious if there is a particular reason why we're using a top-level await here, where on other platforms we didn't?

I'd suggest to either:

 - make it a regular import, like `import wasm from "./pkg/matrix_sdk_crypto_wasm_bg.wasm"`
 - stick it in the `initAsync`

I have a slight preference for the latter for consistency with other entrypoints

package.json
sandhose, line 1: It's missing the new file in the `files` section

sandhose (changes requested):

index-wasm-esm.mjs
richvdh: yeah, I guess the fewer operational differences we have between mjs and cjs the better.

sandhose (approved):

andybalaam (approved):
I can't pretend to understand much of this, but certainly nothing concerning from a Crypto point of view.

PR #179, opened 2024-12-13
feat(dehydration): support for dehydration key and storage

BillCarsonFr:
Fixes https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/160

Depends on https://github.com/matrix-org/matrix-rust-sdk/pull/4399

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,13 @@
     other functions are used. The behaviour is unchanged and still available on
     Node.js.
 
+-   Expose new API `DehydratedDevices.getDehydratedDeviceKey`, `DehydratedDevices.saveDehydratedDeviceKey`
+    and `DehydratedDevices.deleteDehydratedDeviceKey` to store/load the dehydrated device pickle key.
+    This allows client to automatically rotate the dehydrated device to avoid one-time-keys exhaustion and
+    to_device accumulation. `DehydratedDevices.keysForUpload` and `DehydratedDevices.rehydrate` now use the
+    `DehydratedDeviceKey` as parameter instead of a raw UInt8Array.
+    Use `DehydratedDeviceKey::createKeyFromArray` to migrate.
+
 # matrix-sdk-crypto-wasm v11.0.0
 
 -   Update matrix-rust-sdk to `70bcddfba5e19`.
--- a/src/dehydrated_devices.rs
+++ b/src/dehydrated_devices.rs
@@ -3,7 +3,7 @@
 //! WASM wrapper for `matrix_sdk_crypto::dehydrated_devices`.
 
 use js_sys::{Array, JsString, Uint8Array};
-use matrix_sdk_crypto::dehydrated_devices;
+use matrix_sdk_crypto::{dehydrated_devices, store::DehydratedDeviceKey as SdkDehydratedDeviceKey};
 use wasm_bindgen::prelude::*;
 
 use crate::{identifiers::DeviceId, requests::PutDehydratedDeviceRequest, store::RoomKeyInfo};
@@ -21,6 +21,51 @@ impl From<dehydrated_devices::DehydratedDevices> for DehydratedDevices {
     }
 }
 
+/// Dehydrated device key
+#[wasm_bindgen]
+#[derive(Debug)]
+pub struct DehydratedDeviceKey {
+    pub(crate) inner: Uint8Array,
+}
+
+#[wasm_bindgen]
+impl DehydratedDeviceKey {
+    /// Generates a new random pickle key.
+    #[wasm_bindgen(js_name = "createRandomKey")]
+    pub fn create_random_key() -> Result<DehydratedDeviceKey, JsError> {
+        Ok(SdkDehydratedDeviceKey::new()?.into())
+    }
+
+    /// Generates a new random pickle key.
+    #[wasm_bindgen(js_name = "createKeyFromArray")]
+    pub fn create_key_from_array(array: Uint8Array) -> Result<DehydratedDeviceKey, JsError> {
+        Ok(SdkDehydratedDeviceKey::from_slice(array.to_vec().as_slice())?.into())
+    }
+
+    /// Convert the pickle key to a base 64 encoded string.
+    #[wasm_bindgen(js_name = "toBase64")]
+    pub fn to_base64(&self) -> JsString {
+        let binding = self.inner.to_vec();
+        let inner: &[u8; 32] = binding.as_slice().try_into().expect("Expected 32 byte array");
+
+        SdkDehydratedDeviceKey::from(inner).to_base64().into()
+    }
+}
+
+// Zero out on drop
+impl Drop for DehydratedDeviceKey {
+    fn drop(&mut self) {
+        self.inner.fill(0, 0, 32);
+    }
+}
+
+impl From<matrix_sdk_crypto::store::DehydratedDeviceKey> for DehydratedDeviceKey {
+    fn from(pickle_key: matrix_sdk_crypto::store::DehydratedDeviceKey) -> Self {
+        let vec: Vec<u8> = pickle_key.into();
+        DehydratedDeviceKey { inner: vec.as_slice().into() }
+    }
+}
+
 #[wasm_bindgen]
 impl DehydratedDevices {
     /// Create a new [`DehydratedDevice`] which can be uploaded to the server.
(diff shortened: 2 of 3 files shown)


src/dehydrated_devices.rs
```diff
@@ -21,6 +21,51 @@ impl From<dehydrated_devices::DehydratedDevices> for DehydratedDevices {
...
+}
+
+// Zero out on drop
+impl Drop for DehydratedDeviceKey {
+    fn drop(&mut self) {
+        self.inner.fill(0, 0, 32);
+    }
+}
+
+impl From<matrix_sdk_crypto::store::DehydratedDeviceKey> for DehydratedDeviceKey {
```
richvdh, line 62: it's a bit confusing that we use `matrix_sdk_crypto::store::DehydratedDeviceKey` and `SdkDehydratedDeviceKey` within the same file to refer to the same thing. I'd suggest using `SdkDehydratedDeviceKey` throughout, but please pick one and be consistent.

src/dehydrated_devices.rs
```diff
@@ -21,6 +21,51 @@ impl From<dehydrated_devices::DehydratedDevices> for DehydratedDevices {
     }
 }
 
+/// Dehydrated device key
+#[wasm_bindgen]
+#[derive(Debug)]
+pub struct DehydratedDeviceKey {
+    pub(crate) inner: Uint8Array,
```
richvdh, line 28: why not use `SdkDehydratedDeviceKey` here, rather than `Uint8Array`?

richvdh, line 28: Also: does it need to be `pub(crate)`, or could it be private?

src/dehydrated_devices.rs
```diff
@@ -21,6 +21,51 @@ impl From<dehydrated_devices::DehydratedDevices> for DehydratedDevices {
...
+
+#[wasm_bindgen]
+impl DehydratedDeviceKey {
+    /// Generates a new random pickle key.
+    #[wasm_bindgen(js_name = "createRandomKey")]
+    pub fn create_random_key() -> Result<DehydratedDeviceKey, JsError> {
+        Ok(SdkDehydratedDeviceKey::new()?.into())
+    }
+
+    /// Generates a new random pickle key.
```
richvdh, line 39: if we're passing in the array, it's not random?

src/dehydrated_devices.rs
```diff
@@ -21,6 +21,51 @@ impl From<dehydrated_devices::DehydratedDevices> for DehydratedDevices {
...
+    #[wasm_bindgen(js_name = "createKeyFromArray")]
+    pub fn create_key_from_array(array: Uint8Array) -> Result<DehydratedDeviceKey, JsError> {
+        Ok(SdkDehydratedDeviceKey::from_slice(array.to_vec().as_slice())?.into())
+    }
+
+    /// Convert the pickle key to a base 64 encoded string.
```
richvdh, line 45:
```suggestion
    /// Convert the pickle key to a base64-encoded string.
```

src/dehydrated_devices.rs
```diff
@@ -21,6 +21,51 @@ impl From<dehydrated_devices::DehydratedDevices> for DehydratedDevices {
...
+/// Dehydrated device key
+#[wasm_bindgen]
+#[derive(Debug)]
+pub struct DehydratedDeviceKey {
+    pub(crate) inner: Uint8Array,
+}
+
+#[wasm_bindgen]
+impl DehydratedDeviceKey {
+    /// Generates a new random pickle key.
```
richvdh, line 33: I'm a bit confused about why we refer to it as a "pickle key" here (and in the other methods), but the type (and documentation on the type itself) call it a "dehydrated device key".

I guess a "pickle key" and a "dehydrated device key" are the same thing, somehow? Basically it feels like the documentation could do with some clarification. It might be clearer if we stuck with "dehydrated device key" throughout.

src/dehydrated_devices.rs
```diff
@@ -33,18 +78,54 @@ impl DehydratedDevices {
...
             .inner
-            .rehydrate(&pickle_key, &device_id.inner, serde_json::from_str(device_data)?)
+            .rehydrate(&sdk_pickle_key, &device_id.inner, serde_json::from_str(device_data)?)
             .await?
             .into())
     }
+
+    /// Get the cached dehydrated device pickle key if any.
+    ///
+    /// None if the key was not previously cached (via
```
richvdh, line 97:
```suggestion
    /// `None` if the key was not previously cached (via
```

src/dehydrated_devices.rs
```diff
@@ -33,18 +78,54 @@ impl DehydratedDevices {
...
+    /// [`Self::save_dehydrated_device_pickle_key`]).
+    ///
+    /// Should be used to periodically rotate the dehydrated device to avoid
+    /// OTK exhaustion and accumulation of to_device messages.
+    #[wasm_bindgen(js_name = "getDehydratedDeviceKey")]
+    pub async fn get_dehydrated_device_key(&self) -> Result<Option<DehydratedDeviceKey>, JsError> {
+        let key = self.inner.get_dehydrated_device_pickle_key().await?;
+        Ok(key.map(DehydratedDeviceKey::from))
```
richvdh, line 105: maybe simpler:

```suggestion
        Ok(key.into())
```

src/dehydrated_devices.rs
```diff
@@ -33,18 +78,54 @@ impl DehydratedDevices {
...
-            .rehydrate(&pickle_key, &device_id.inner, serde_json::from_str(device_data)?)
+            .rehydrate(&sdk_pickle_key, &device_id.inner, serde_json::from_str(device_data)?)
             .await?
             .into())
     }
+
+    /// Get the cached dehydrated device pickle key if any.
+    ///
+    /// None if the key was not previously cached (via
+    /// [`Self::save_dehydrated_device_pickle_key`]).
```
richvdh, line 98: is it correct that `save_dehydrated_device_pickle_key` is the only way to save the key? It is not loaded automatically (eg via secret sharing) at any point?

src/dehydrated_devices.rs
```diff
@@ -33,18 +78,54 @@ impl DehydratedDevices {
...
+    ///
+    /// Should be used to periodically rotate the dehydrated device to avoid
+    /// OTK exhaustion and accumulation of to_device messages.
```
richvdh, line 101: The way this is written, it sounds like calling this method will rotate the dehydrated device, which I don't think is correct?

src/dehydrated_devices.rs
```diff
@@ -33,18 +78,54 @@ impl DehydratedDevices {
...
+    #[wasm_bindgen(js_name = "getDehydratedDeviceKey")]
+    pub async fn get_dehydrated_device_key(&self) -> Result<Option<DehydratedDeviceKey>, JsError> {
+        let key = self.inner.get_dehydrated_device_pickle_key().await?;
+        Ok(key.map(DehydratedDeviceKey::from))
+    }
+
+    /// Store the dehydrated device pickle key in the crypto store.
+    ///
+    /// This is useful if the client wants to periodically rotate dehydrated
+    /// devices to avoid OTK exhaustion and accumulated to_device problems.
```
richvdh, line 111: I think I need a bit more explanation about how to use this. Do I just make a new `DehydratedDeviceKey` and call `save_dehydrated_device_key` with it?

richvdh (changes requested):
A few questions and requests for clarification

src/dehydrated_devices.rs
BillCarsonFr: Yes that's what I wanted to do intitialy. But there was an issue doing that with ffi bingings. Let me try with wasm

src/dehydrated_devices.rs
BillCarsonFr: Right tried to be consistent with other From<> impl. but will use the same.
Also there is some inconsistency in the was we name the import alias: sometimes it is `SdkThing` , `InnerThing` or `RustThing`. Let me try to clean that a bit

BillCarsonFr: oh I can see `OriginalThing` too

src/dehydrated_devices.rs
BillCarsonFr: sry. bsd copy paste

src/dehydrated_devices.rs
BillCarsonFr: I don't have a From<Option<>> , not sure to get it?

src/dehydrated_devices.rs
BillCarsonFr: this part is not yet done.

src/dehydrated_devices.rs
BillCarsonFr: no it doesn't rotate anything. Maybe should be `usefull` ?

src/dehydrated_devices.rs
BillCarsonFr: Maybe I should remove these comments if they are bringing confusion?

The rotation of dehydrated device is another topic not handled by this code. But caching the key and re-use it is important if you want to rotate without having to ask for the 4S key every time (in order to update 4S with a new dehydrated device key for the next login to use)

src/dehydrated_devices.rs
richvdh: sorry, you're right, this doesn't work.

src/dehydrated_devices.rs
richvdh: Ok. We'll need to remember to update this comment when this changes.

src/dehydrated_devices.rs
```diff
@@ -33,18 +71,53 @@ impl DehydratedDevices {
...
+                serde_json::from_str(device_data)?,
+            )
             .await?
             .into())
     }
+
+    /// Get the cached dehydrated device key if any.
+    ///
+    /// `None` if the key was not previously cached (via
+    /// [`DehydratedDevices::save_dehydrated_device_key`]).
```
richvdh, line 92: oh, another thing: this style of link doesn't work for the wasm bindings. You have to use `{@link ...}`, as we do for typescript.

Unfortunately we haven't wired up generating the documentation in CI. Could you generate it locally (`yarn doc`) and check the results?

src/dehydrated_devices.rs
richvdh: hrm; I think it would be *nice* to have documentation of how to correctly rotate a dehydrated device. But it's not essential, and yes, this comment is probably bringing more confusion than help.

On balance: yes, probably best to remove this comment.

src/dehydrated_devices.rs
uhoreg: Maybe something like "Should be used so that the dehydrated device can be rotated ..."?

src/dehydrated_devices.rs
BillCarsonFr: Using `dehydrated device key` everywhere now.
I'll have to do something similar in sdk and ffi

src/dehydrated_devices.rs
BillCarsonFr: looks like I had a bad understanding of pub(crate). Thx

src/dehydrated_devices.rs
BillCarsonFr: Thx fixed this one and another in this file

Cargo.lock
```diff
@@ -1044,8 +1045,8 @@ dependencies = [
 
 [[package]]
 name = "matrix-sdk-common"
-version = "0.8.0"
-source = "git+https://github.com/matrix-org/matrix-rust-sdk?branch=release-for-crypto-wasm-12#37c17cf854a70fe6f719c7fde49a6b0e4402988f"
+version = "0.9.0"
+source = "git+https://github.com/matrix-org/matrix-rust-sdk#b18100228e0456973d56d4ac6a58caf30b17bd8d"
```
andybalaam, line 1049: This will have a conflict with https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/184 but you can just keep your changes and ignore that.

andybalaam (approved):
Looks good, thanks.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # UNRELEASED
 
+**BREAKING CHANGES**
+
+-   Update matrix-rusk-sdk to `0.9.0`.
```
BillCarsonFr, line 5: @andybalaam Should I use a git hash here too? even it is this actual tag, or link the tag?

andybalaam: I think what you have is sufficient.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+-   Expose new API `DehydratedDevices.getDehydratedDeviceKey`, `DehydratedDevices.saveDehydratedDeviceKey`
+    and `DehydratedDevices.deleteDehydratedDeviceKey` to store/load the dehydrated device pickle key.
+    `DehydratedDevices.keysForUpload` and `DehydratedDevices.rehydrate` now use the `DehydratedDeviceKey` as parameter
+    instead of a raw UInt8Array.Use `DehydratedDeviceKey::createKeyFromArray` to migrate.
```
richvdh, line 9: Sorry, a belated review:

* I think this point about `keysForUpload` and `rehydrate` should be a separate bullet point
* The other points ("Update matrix-rusk-sdk to `0.9.0`" and "Expose new API") aren't breaking changes so should be *above* the "**BREAKING CHANGES**" header (see matrix-sdk-crypto-wasm v12.0.0, for example)
* It would be good to link to this pr for the changes (see lots of other entries in the changelog, for example)
* There are a number of typos on line 9. It should be: 
   ```
       instead of a raw `UInt8Array`. Use `DehydratedDeviceKey.createKeyFromArray` to migrate.
   ```

PR #184, opened 2024-12-19
Go back to depending on matrix-rust-sdk main branch

andybalaam:
This undoes the change made in 0378300821d2aeca7d2f184141c864f2fc189051, except it stays on the same commit SHA of matrix-rust-sdk, but changes the branch specifier.

We temporarily switched to the branch `release-for-crypto-wasm-12` of `matrix-rust-sdk` to release 12.1.0 without needing to fix the other things needing work to get the latest `matrix-rust-sdk`. This change puts us back onto the main branch, as normal.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1045,7 +1045,7 @@ dependencies = [
 [[package]]
 name = "matrix-sdk-common"
 version = "0.8.0"
-source = "git+https://github.com/matrix-org/matrix-rust-sdk?branch=release-for-crypto-wasm-12#37c17cf854a70fe6f719c7fde49a6b0e4402988f"
+source = "git+https://github.com/matrix-org/matrix-rust-sdk#37c17cf854a70fe6f719c7fde49a6b0e4402988f"
 dependencies = [
  "async-trait",
  "eyeball-im",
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -1045,7 +1045,7 @@ dependencies = [
 [[package]]
 name = "matrix-sdk-common"
 version = "0.8.0"
-source = "git+https://github.com/matrix-org/matrix-rust-sdk?branch=release-for-crypto-wasm-12#37c17cf854a70fe6f719c7fde49a6b0e4402988f"
+source = "git+https://github.com/matrix-org/matrix-rust-sdk#e99939db857ca36054c0023ef68cc181236fdee4"
```
richvdh, line 1048: can we do this without rolling back the rust-sdk? ie, just:
```suggestion
source = "git+https://github.com/matrix-org/matrix-rust-sdk#37c17cf854a70fe6f719c7fde49a6b0e4402988f"
```

(likewise below)

andybalaam: Good idea. I force-pushed since adding a commit would be more confusing imo.

richvdh (approved):
LGTM

PR #189, opened 2025-01-08
Fix splitbrain between `.js` being ESM & CJS inconsistently

t3chguy:
Fixes https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/188

no actual changes to the `js/cjs/mjs` files barring some ts annotations

Switches the package type to ESM as that is what wasm-pack is outputting
Convinces the wasm bindgen test to use an ESM runner
Uses explicit extensions on the CJS entrypoints


richvdh:
would it be better to stop using `.js` altogether, and just call everything `.mjs` / `.cjs`, to avoid this sort of confusion?

t3chguy:
That would also be possible but I somehow hate the idea of having to rename the output files of wasm-pack `s/js/mjs` - `"type": "module"` is also a good thing going forward. I can imagine also imagine issues going forward if the tests/entrypoints were ts-ified and needing `mts/cts` and the issues typescript has around those

andybalaam:
@richvdh I assigned this to you since you seem to understand it more than me. Feel free to assign another one to me :-)

node.cjs
```diff
@@ -59,7 +59,7 @@ let initialised = false;
  *
  * It will throw if there is an attempt to load the module asynchronously running
  *
- * @returns {typeof import("./pkg/matrix_sdk_crypto_wasm_bg.wasm.d")}
+ * @returns {typeof import("./pkg/matrix_sdk_crypto_wasm_bg.wasm.d", { with: { "resolution-mode": "require" } })}
```
richvdh, line 62: Can you help me understand what exactly this is doing? I get that it's to do with https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-3.html?#stable-support-resolution-mode-in-import-types, but what's the actual outcome/why do we need this?

richvdh (commented):
> no actual changes to the js/cjs/mjs files

This doesn't seem quite right? Could you give the description of the PR an update, please, to give an overview of the approach? (Looks like we're now standardising on `.js` meaning ESM, and `.cjs` meaning CommonJS?)

Also, please can you add an entry to the changelog?

node.cjs
t3chguy: image

Its what the error told me to do :D

t3chguy:
> This doesn't seem quite right? Could you give the description of the PR an update, please, to give an overview of the approach? (Looks like we're now standardising on .js meaning ESM, and .cjs meaning CommonJS?)

Have updated the description

> Also, please can you add an entry to the changelog?

Could you please point me to any guidance on doing so? Not seeing anything in the contributing doc

richvdh:
>> Also, please can you add an entry to the changelog?
> Could you please point me to any guidance on doing so? Not seeing anything in the contributing doc

https://github.com/matrix-org/matrix-rust-sdk/blob/main/CONTRIBUTING.md#writing-changelog-entries is the most relevant guide.

TL;DR though: if you could just add an entry under the "UNRELEASED" heading, following the same style as previous entries, that would be be great.

node.cjs
richvdh: Hrm, fair enough. I assume the commandline `yarn lint:types`/`tsc --noEmit` showed the same error? Or was this specific to your IDE?

I'd be interested to understand why this is a problem now when it wasn't before.

tsc is now detecting `node.cjs` as CommonJs where it wasn't before? (which seems odd, why would it have done that?) 

Or tsc now considers `./pkg/matrix_sdk_crypto_wasm_bg.wasm.d.ts` as ESMish where it wasn't before? (seems odd, it's typescript)

t3chguy: > Hrm, fair enough. I assume the commandline yarn lint:types/tsc --noEmit showed the same error? Or was this specific to your IDE?

Same error

```
/opt/homebrew/bin/yarn run lint:types
yarn run v1.22.22
$ tsc --noEmit
node.cjs:83:36 - error TS1542: Type import of an ECMAScript module from a CommonJS module must have a 'resolution-mode' attribute.

83  * @returns {Promise<typeof import("./pkg/matrix_sdk_crypto_wasm_bg.wasm.d")>}
                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in node.cjs:83

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

Process finished with exit code 2
```

> Or tsc now considers ./pkg/matrix_sdk_crypto_wasm_bg.wasm.d.ts as ESMish where it wasn't before? (seems odd, it's typescript)

Yes exactly that, because `.mts` and `.cts` exist, `.ts` esm/cjs also depends on package.json `type` field.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # UNRELEASED
 
+-   The published package now uses type=module which defaults js/ts files to ESM.
```
richvdh, line 3: I feel like this doesn't really tell me what I need to know as a user of this library.

```suggestion
-   Fix a problem, introduced in v12.0.0, when importing the published package as an ESM module, in which some files could be incorrectly interpreted as CommonJS, leading to syntax errors.
    ([#189](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/189))
```

Also, your branch is outdated. Please merge and resolve conflicts.

richvdh (approved):
LGTM. Thank you.

PR #193, opened 2025-01-20
Remove uses of `future_to_promise` in `machine.rs`

richvdh:
Modern wasm_bindgen allows us just to write `async` functions, which is much more idiomatic, and gives better typescript types in the generated output.


Hywan (approved):
Very nice!

andybalaam (approved):
Woohoo!

PR #194, opened 2025-01-28
Update dependency versions

richvdh:
`cargo update`, which brings in https://github.com/matrix-org/matrix-rust-sdk/pull/4581.

Fixes: https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/180


andybalaam (approved):
Looks good. I personally prefer to update matrix-rust-sdk in a different commit from the general deps updates.

PR #205, opened 2025-02-17
Revert "Remove uses of `future_to_promise` in `machine.rs` (#193)"

richvdh:
This reverts commit eea167e45470cfa387a89ea5b4b04e2a579891b5.

Unfortunately, most of this change is unsound.

With an explicit `future_to_promise`, we are able to clone the underlying `matrix_sdk_crypto::OlmMachine` (and its constituent `Arc`) *before* returning a Promise to the javascript, meaning even if the wrapping `matrix_sdk_crypto_wasm::OlmMachine` is dropped, the underlying `OlmMachine` hangs around.

An `async` function does nothing until it is polled, meaning there is plenty of time for the Javascript to get hold of the Promise and drop its reference to the OlmMachine, before we try and clone it. It's fine (I think) if we let the Javascript finalizer manage cleanup; but we deliberately drop OlmMachine, in an attempt to get it to close its indexeddb connection.

I'd go so far as to say that wasm-bindgen's support for `#[wasm_bindgen]` on `async fn` (https://github.com/rustwasm/wasm-bindgen/pull/1754) is fundamentally incompatible with an explicit call to `.free()` (or, in this case, `.close()`, which is the same thing.)

Probably, the correct fix here is to improve the way we handle indexeddb connections (cf https://github.com/element-hq/element-web/issues/25779), at which point we can remove the `.close()`/`.free()` calls, and reintroduce this. For now though, I'm going to back it out.

Fixes: https://github.com/matrix-org/matrix-rust-sdk/issues/4672

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # UNRELEASED
 
+-   Revert updates to generated typescript types, which could cause runtime
+    errors.
+    ([#205](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/205))
+
 # matrix-sdk-crypto-wasm v14.0.0
 
 -   `CollectStrategy.deviceBasedStrategy` is deprecated, and replaced by other methods in `CollectStrategy`.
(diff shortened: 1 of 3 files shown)


Hywan (approved):
Good to me. Thanks for catching this!

PR #206, opened 2025-02-17
Reintroduce stricter typescript types

richvdh:
A partner to #205: the reverted PR gave some better typescript types. Rather than lose them altogether, we can replicate them manually.

A second commit adds a few more improvements.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -47,3 +47,56 @@ export * from "./pkg/matrix_sdk_crypto_wasm.js";
  * @returns {Promise<void>}
  */
 export declare function initAsync(): Promise<void>;
+
+// The auto-generated typescript definitions are a good start, but could do with tightening up in a lot of areas.
+// The following is a manually-curated set of typescript definitions.
+declare module "./pkg/matrix_sdk_crypto_wasm.js" {
+    interface OlmMachine {
+        trackedUsers(): Promise<Set<any>>;
+        updateTrackedUsers(users: UserId[]): Promise<void>;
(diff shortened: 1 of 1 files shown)


bnjbvr (approved):
I will approve any PR that adds more typing \o/ Thanks. I assume that if the CI passes, then the code type-checks against these new types, and I don't need to thoroughly review each type in signatures, but let me know if that's not the case.

richvdh:
> I will approve any PR that adds more typing \o/ Thanks. I assume that if the CI passes, then the code type-checks against these new types, and I don't need to thoroughly review each type in signatures, but let me know if that's not the case.

Unfortunately, I don't think it's the case that we have much in the way of CI that checks these types are actually correct. I don't necessarily think that means we need to go through and carefully review the declarations one-by-one though.

Firstly, I'd argue that, even if the types are wrong, the code is no more dangerous than it was before when we had no type declarations at all.

Secondly, I have compile-tested this against matrix-js-sdk, which makes more extensive use of the library and is likely to highlight any grave mistakes.

So, up to you if you want to do a more rigorous review. If you do, I recommend checking the first commit ([268ebc7](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/206/commits/268ebc78992178005056b07f1bf9021992ad5ec8)) against the autogenerated types in v14.0.0 (https://www.runpkg.com/?@matrix-org/matrix-sdk-crypto-wasm@14.0.0/pkg/matrix_sdk_crypto_wasm.d.ts); the second commit will need more manual verification.

bnjbvr:
I trust you and usage in the JS SDK, thanks for the summary!

PR #220, opened 2025-04-15
Update dependencies, other than matrix-rust-sdk

richvdh:
Update all of our dependencies, other than matrix-rust-sdk.

Updating `ulid` and `vergen` is not completely trivial, so they are pulled out to separate commits.

There was previously a problem with serde-wasm-bindgen (https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/161), but it seems that has been fixed.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,5 @@
 [build]
 target = "wasm32-unknown-unknown"
+
+[target.wasm32-unknown-unknown]
+rustflags = ['--cfg', 'getrandom_backend="wasm_js"']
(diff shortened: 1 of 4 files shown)


.cargo/config.toml
```diff
@@ -1,2 +1,5 @@
 [build]
 target = "wasm32-unknown-unknown"
+
+[target.wasm32-unknown-unknown]
+rustflags = ['--cfg', 'getrandom_backend="wasm_js"']
```
Hywan, line 5: Smart to put that here 👌.

Hywan (approved):
Looks good to me, thanks!

.cargo/config.toml
richvdh: I can't claim too much credit. I just read the `getrandom` docs and did what it told me...

PR #226, opened 2025-05-09
Support ProcessedToDeviceEvent variants returned by receiveSyncChanges

BillCarsonFr:
Draft for now as I will check if a rust-sdk release is underway (to avoid pinning to a random commit)


CHANGELOG.md
```diff
@@ -1,8 +1,17 @@
 # UNRELEASED
 
+-   Update matrix-rusk-sdk to `35a2ce9`, which includes:
+
+    -   Add variants for plain text and encrypted to-device events ([#4935](https://github.com/matrix-org/matrix-rust-sdk/pull/4935))
```
richvdh, line 5: Does this affect the API exposed by crypto-wasm? If so, it would be good to say how. If not, it would probably be good to say so explicitly.

BillCarsonFr:
@richvdh As discussed I have updated to the latest rust change to include https://github.com/matrix-org/matrix-rust-sdk/pull/5074

There have been other changes on the sdk, to wrap things around `Box` and `Arc`, I think the changes are fine but you might want to look at this commit for this is https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/226/commits/dfa97e8f4fad112c9749407fed307c0bdb6d2f7e

I then changed `ProcessedToDeviceEvent` mapping to JS using the same pattern we use for `OutgoingRequest` https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/226/commits/df25e93dac2d14c61fc2a4e341b69bca91baef35

richvdh (changes requested):
A few comments and suggestions.

Please could you also update the title and description of the PR to be a more accurate representation of what is changing here.

src/types.rs
```diff
@@ -363,123 +364,139 @@ impl From<&RoomSettings> for matrix_sdk_crypto::store::RoomSettings {
...
 
-    /// An invalid to device event that was ignored because it is missing some
+    /// An invalid to-device message that was ignored because it is missing some
     /// required information to be processed (like no event `type` for
     /// example)
     Invalid,
 }
 
-/// Represent a ToDevice event after it has been decrypted by {@link
+/// Represent a to-device event after it has been decrypted by {@link
```
richvdh, line 386: Still lots of instances of "Represent ..." in this PR.

src/types.rs
```diff
@@ -363,123 +364,139 @@ impl From<&RoomSettings> for matrix_sdk_crypto::store::RoomSettings {
...
+    #[wasm_bindgen(getter, js_name = "type")]
+    pub fn processed_type(&self) -> ProcessedToDeviceEventType {
+        ProcessedToDeviceEventType::Decrypted
+    }
+
+    /// Create a new `DecryptedToDeviceEvent`.
+    #[wasm_bindgen(constructor)]
+    pub fn new(decrypted_raw_event: JsString, encryption_info: OlmEncryptionInfo) -> Self {
+        Self { decrypted_raw_event, encryption_info }
+    }
```
richvdh, line 417: it's a bit odd to put the constructor *after* methods. Could you you put the constructor before `processed_type`, please.

src/types.rs
```diff
@@ -363,123 +364,139 @@ impl From<&RoomSettings> for matrix_sdk_crypto::store::RoomSettings {
...
-    /// A JSON-encoded string containing the processed event.
-    /// The decrypted event as if it was sent in clear (For room keys or secrets
-    /// some part of the content might have been zeroize'd).
-    #[wasm_bindgen(readonly, js_name = "wireEvent")]
-    pub wire_event: JsString,
+    /// A JSON-encoded object containing the processed event.
+    /// The decrypted message as if it was sent in clear (For room keys or
+    /// secrets some part of the content might have been zeroize'd).
+    /// Represents a Matrix to-device message with `type`, `sender` and
+    /// `content` fields.
```
richvdh, line 395:
```suggestion
    /// A JSON-encoded object containing the decrypted event, as if it had been
    /// sent in the clear. 
    ///
    /// Typically contains properties `type`, `sender` and `content`.
    ///
    /// (For room keys or secrets, some part of the content might have been zeroized).
```

src/responses.rs
```diff
@@ -299,15 +309,67 @@ impl EncryptionInfo {
...
+    /// delayed (for room keys for example). For custom to-devices there is
+    /// no delay, this will not be None (the decryption would fail if the
+    /// sender device keys cannot be found).
+    /// Note this is untrusted data unless `isSenderVerified` is also true.
+    #[wasm_bindgen(getter, js_name = "senderDevice")]
+    pub fn sender_device(&self) -> Option<identifiers::DeviceId> {
+        Some(self.inner.sender_device.as_ref()?.clone().into())
+    }
+
+    /// The Curve25519 key of the device that encrypted the message.
```
richvdh, line 356:
```suggestion
    /// The base64-encoded public Curve25519 key of the device that encrypted the message.
```

src/responses.rs
```diff
@@ -299,15 +309,67 @@ impl EncryptionInfo {
...
+    #[wasm_bindgen(getter, js_name = "senderCurve25519Key")]
+    pub fn sender_curve25519_key(&self) -> Option<JsString> {
+        match &self.inner.algorithm_info {
+            AlgorithmInfo::MegolmV1AesSha2 { .. } => {
+                panic!("MegolmV1AesSha2 variant cannot be converted to OlmEncryptionInfo")
+            }
+            AlgorithmInfo::OlmV1Curve25519AesSha2 { curve25519_public_key_base64 } => {
+                Some(curve25519_public_key_base64.clone().into())
             }
         }
```
richvdh, line 366: it would probably be better to copy the `curve25519_public_key_base64` out of `algorithm_info` in `TryFrom<matrix_sdk_common::deserialized_responses::EncryptionInfo>`, to avoid this test and `panic`.

And since you're doing that, maybe consider copying `sender`, `verification_state`, etc.

src/responses.rs
```diff
@@ -299,15 +309,67 @@ impl EncryptionInfo {
...
+#[wasm_bindgen()]
+impl OlmEncryptionInfo {
+    /// The user ID of the to-device message sender. Note this is untrusted data
+    /// unless `isSenderVerified` is also true.
+    #[wasm_bindgen(getter)]
+    pub fn sender(&self) -> identifiers::UserId {
+        identifiers::UserId::from(self.inner.sender.clone())
+    }
+
+    /// The device ID of the device that sent us the to-device message.
```
richvdh, line 345:
```suggestion
    /// The device ID of the device that sent us the to-device message.
    ///
```

src/responses.rs
```diff
@@ -299,15 +309,67 @@ impl EncryptionInfo {
...
+    /// Could be None in the case where the to-device message sender checks are
+    /// delayed (for room keys for example). For custom to-devices there is
```
richvdh, line 347:
```suggestion
    /// delayed (for room keys for example). For custom to-device messages there is
```

richvdh: what exactly is a "custom" to-device message here?

src/responses.rs
```diff
@@ -299,15 +309,67 @@ impl EncryptionInfo {
...
+    /// no delay, this will not be None (the decryption would fail if the
```
richvdh, line 348:
```suggestion
    /// no delay, so this will not be `undefined` (the decryption would fail if the
```

src/responses.rs
```diff
@@ -299,15 +309,67 @@ impl EncryptionInfo {
...
+    /// sender device keys cannot be found).
+    /// Note this is untrusted data unless `isSenderVerified` is also true.
```
richvdh, line 350:
```suggestion
    /// sender device keys cannot be found).
    ///
    /// Note this is untrusted data unless `isSenderVerified` is also true.
```

src/types.rs
```diff
@@ -494,20 +511,21 @@ pub fn processed_to_device_event_to_js_value(
...
-            JsValue::from(DecryptedToDeviceEvent {
-                processed_type: ProcessedToDeviceEventType::Decrypted,
-                wire_event: raw.json().get().to_owned().into(),
-                encryption_info: encryption_info.into(),
-            })
+            JsValue::from(DecryptedToDeviceEvent::new(
+                raw.json().get().into(),
+                encryption_info
+                    .try_into()
+                    .expect("to-device encryption info will always convert to OlmEncryptionInfo"),
```
richvdh, line 518: it might be good to explain in more detail why this is the case.

I think it might be better just to drop such events. (Easiest way to do that: have `processed_to_device_event_to_js_value` return `Option<JsValue>` and use `filter_map` instead of `map` at the call site.)

src/responses.rs
```diff
@@ -216,19 +216,34 @@ impl From<matrix_sdk_common::deserialized_responses::TimelineEvent> for Decrypte
     fn from(value: matrix_sdk_common::deserialized_responses::TimelineEvent) -> Self {
         Self {
             event: value.raw().json().get().to_owned().into(),
-            encryption_info: value.encryption_info().map(|e| e.clone().into()),
+            encryption_info: value.encryption_info().map(|e| {
+                e.as_ref().clone().try_into().expect("Timeline events are megolm encrypted")
+            }),
         }
     }
```
richvdh, line 223: I'd probably make this into try_from rather than panic.

src/responses.rs
```diff
@@ -251,9 +266,12 @@ impl EncryptionInfo {
...
     pub fn sender_curve25519_key(&self) -> Option<JsString> {
-        Some(match &self.inner.algorithm_info {
-            AlgorithmInfo::MegolmV1AesSha2 { curve25519_key, .. } => curve25519_key.clone().into(),
-        })
+        match &self.inner.algorithm_info {
+            AlgorithmInfo::MegolmV1AesSha2 { curve25519_key, .. } => {
+                Some(curve25519_key.clone().into())
+            }
+            AlgorithmInfo::OlmV1Curve25519AesSha2 { .. } => None,
+        }
```
richvdh, line 274: again, consider copying the `curve25519_key` and `sender_claimed_ed25519_key` into the `EncryptionInfo` struct, to avoid having to do this match.

Alternatively, add a comment saying that we know the algorithm is MegolmV1AesSha2

richvdh (changes requested):

richvdh:
I'm actually going to split this out into smaller PRs.

PR #228, opened 2025-05-21
Log warnings when we fail to parse a backed-up room key

andybalaam:
Would have helped when debugging https://github.com/element-hq/element-web/issues/29878

Errors like this were made non-fatal in https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/85 but here we make them a little more visible.

Please review this carefully as I don't have any way to write tests for this code (right?).


Hywan (approved):
LGTM, thanks.

PR #235, opened 2025-06-10
Combine the 'test' and 'docs' GHA jobs

richvdh:
We have two GHA jobs that both spend ages building everything. Won't someone think of the polar bears?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,11 @@ jobs:
     name: test
     runs-on: ubuntu-latest
 
+    permissions:
+      contents: read
+      pages: write
+      id-token: write
+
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -90,54 +95,15 @@ jobs:
...
-      id-token: write
-
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v3
-
-      - name: Install Protoc
-        uses: arduino/setup-protoc@v1
-        with:
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
richvdh, line 115: I have no idea why we installed this. We don't use it afaict.

PR #238, opened 2025-06-18
Bump sdk to 022068996

richvdh:
Update the wasm bindings as far as https://github.com/matrix-org/matrix-rust-sdk/commit/022068996.

The breaking changes here are https://github.com/matrix-org/matrix-rust-sdk/pull/5048 and https://github.com/matrix-org/matrix-rust-sdk/pull/5134, which wrap things in `Box` or `Arc`. Review commit-by-commit.

The second of several PRs that I am pulling out of #226.

PR #239, opened 2025-06-18
Expose EncryptionInfo on decrypted to-device events

richvdh:
Bump SDK to https://github.com/matrix-org/matrix-rust-sdk/commit/0f73ffde, which exposes encryption info on to-device events.

We add a new `ToDeviceEncryptionInfo` struct, which is a member of  `DecryptedToDeviceEvent`.

Review commit-by-commit.

Based on https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/238.

The third of several PRs that I am pulling out of #226.

### DIFF
--- a/src/responses.rs
+++ b/src/responses.rs
@@ -230,7 +230,7 @@ impl TryFrom<matrix_sdk_common::deserialized_responses::TimelineEvent> for Decry
     }
 }
 
-/// Struct containing information on how an event was decrypted.
+/// Struct containing information on how a room event was decrypted.
 #[wasm_bindgen()]
 #[derive(Debug)]
 pub struct EncryptionInfo {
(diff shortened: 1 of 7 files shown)


src/responses.rs
```diff
@@ -312,3 +312,66 @@ impl TryFrom<Arc<matrix_sdk_common::deserialized_responses::EncryptionInfo>> for
...
+    #[wasm_bindgen(getter_with_clone)]
+    pub sender: identifiers::UserId,
+
+    /// The device ID of the device that sent us the to-device message.
+    ///
+    /// Could be `undefined` in the case where the to-device message sender
+    /// checks are delayed. There is no delay for to-device messages other
+    /// than `m.room_key`, so this will not always be truthy for other
+    /// message types (the decryption would fail if the sender device keys
+    /// cannot be found).
```
andybalaam, line 337: I'm having trouble understanding this. Should the "not" be there?

src/types.rs
```diff
@@ -476,12 +482,31 @@ impl InvalidToDeviceEvent {
 /// JavaScript has no complex enums like Rust. To return structs of
 /// different types, we have no choice other than hiding everything behind a
 /// `JsValue`.
+///
+/// If the event cannot be mapped into one of the types we return, returns
+/// `None`.
```
andybalaam, line 487: Just "we return `None`" or "returns `None`"?

andybalaam (approved):
Looks good with a couple of questions from me.

src/responses.rs
richvdh: no, good catch, thanks.

PR #241, opened 2025-06-19
Drop dependency on `anyhow`

richvdh:
Preparation for upgrading to matrix-rust-sdk 0.12.0.

As of https://github.com/matrix-org/matrix-rust-sdk/pull/5082, some of the the errors returned by the rust SDK are no longer `Send + Sync`. That means we can no longer pass them into `anyhow::Error::from()`.

As it turns out, most of our uses of `anyhow::Error` are redundant anyway. This patch set (review commit-by-commit) removes the places we use `anyhow::Error`, and means that we will be able to build against rust-sdk 0.12.0.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1253,6 +1253,7 @@ dependencies = [
  "serde",
  "serde-wasm-bindgen",
  "serde_json",
+ "thiserror 2.0.12",
  "tracing",
  "tracing-subscriber",
  "url",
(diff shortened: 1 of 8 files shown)


andybalaam (approved):

richvdh:
Not entirely sure this is necessary now that https://github.com/matrix-org/matrix-rust-sdk/pull/5256 has landed, but I think it's a net improvement anyway.

PR #242, opened 2025-06-19
Fix nightly sdk compatibility build, again

richvdh:
The previous attempt to fix the nightly `latest-matrix-sdk-crypto` job failed,
because the version of the rust-sdk in git was 0.12.0, whereas our Cargo.lock
required 0.11.0.

It seems like we have to edit the Cargo.toml. Editing a toml isn't something I
want to do from a script, so I've made it an xtask.


.cargo/config.toml
```diff
@@ -1,5 +1,5 @@
-[build]
-target = "wasm32-unknown-unknown"
```
richvdh, line 2: We have to drop this, otherwise the xtask gets built as `wasm32-unknown-unknown`, and therefore won't run.

It was only really here as a convenience for people opening the project in their IDE anyway. There are better ways to deal with that.

xtask/.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
andybalaam, line 1: Should this `.idea` be gitignored?

xtask/src/main.rs
```diff
@@ -0,0 +1,81 @@
...
+fn update_cargo_toml() -> Result<()> {
+    let cargo_toml = "Cargo.toml";
+
+    let mut doc: DocumentMut = fs::read_to_string(cargo_toml)?.parse()?;
+
+    let dependencies = doc["dependencies"].as_table_mut().expect("'dependencies' not a table");
+
+    // Search for dependencies whose name starts 'matrix-sdk', and edit them
+    let mut modified = false;
+    for (name, dep) in dependencies.iter_mut().filter(|(name, _)| name.starts_with("matrix-sdk-")) {
```
andybalaam, line 58: Would be great to unit test this with simple before and after of a Cargo.toml, and hopefully not too hard? I'd probably not bother testing reading/writing to a file etc. but I would prefer a test of the actual modifications.

andybalaam (approved):
One recommended but not mandatory suggestion.

Thanks for looking into this.

xtask/src/main.rs
richvdh: fair

andybalaam (approved):
Thanks!

PR #243, opened 2025-06-20
Drop conversion of DecryptedRoomEvent into TimelineEvent

richvdh:
https://github.com/matrix-org/matrix-rust-sdk/commit/5a1909aab9f70123ad05c37c1305270a6de439fe removes the implicit converstion from `DecryptedRoomEvent` to `TimelineEvent` that we'd been relying on since d7eca6fad (which in turn was due to https://github.com/matrix-org/matrix-rust-sdk/commit/1d1863d323842274b8efaf9aff5542a91a81d68d).

It turns out we don't really need to go via TimelineEvent. We can strip out the conversion, and then strip out some redundant `Option`s.

Review commit-by-commit.


src/responses.rs
```diff
@@ -158,38 +158,38 @@ pub struct DecryptedRoomEvent {
...
     pub event: JsString,
 
-    encryption_info: Option<EncryptionInfo>,
+    encryption_info: EncryptionInfo,
 }
 
 #[wasm_bindgen]
 impl DecryptedRoomEvent {
     /// The user ID of the event sender, note this is untrusted data
     /// unless the `verification_state` is as well trusted.
```
andybalaam, line 167: If you can be bothered, it would be nice to re-word this and the similar docs in this file

andybalaam (approved):
Nice, thanks.

PR #260, opened 2025-08-29
feat: Add experimental encrypted state event support.

kaylendog:
Introduces support for experimental state event encryption, mostly conforming to https://github.com/matrix-org/matrix-rust-sdk/issues/5397. Features are **NOT** locked behind any feature flags, but this should not cause issues as we use `Self::default()` in the marked constructor.

- Adds `OlmMachine.encryptStateEvent`;
- Adds `EncryptionSettings.encryptStateEvents`.
- Adds `RoomSettings.encryptStateEvents`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # UNRELEASED
 
+- Add support for experimental encrypted state events, introduce `EncryptionSettings::encrypt_state_events`,
+  `RoomSettings::encrypt_state_events`, and `OlmMachine::encrypt_state_event`.
+  ([#260](https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/260/files))
+
 # matrix-sdk-crypto-wasm v15.2.0
 
 -   Update matrix-rust-sdk to 0.14.0, which includes:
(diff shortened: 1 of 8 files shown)


Hywan (commented):
Would be awesome to get a test. Is it possible?

richvdh (changes requested):
oops needs changelog pls

richvdh:
> Features are locked behind `experimental-encrypted-state-events`.

this is no longer correct and could do with clarification?

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # UNRELEASED
 
+-   **BREAKING** Add support for experimental encrypted state events, introduce
+    `EncryptionSettings::encrypt_state_events`, `RoomSettings::encrypt_state_events`,
+    and `OlmMachine::encrypt_state_event`.
```
richvdh, line 5: This is predominantly a javascript library, so these things have different names as seen by a user of the library (`EncryptionSettings.encryptStateEvents`, etc).

For the same reason: I don't think this is actually a breaking change: as you demonstrate in your test, `new EncryptionSettings()` continues to work just fine.

richvdh (approved):
LGTM!

PR #269, opened 2025-10-02
feat: Support room key bundle withheld info.

kaylendog:
Upstream PR in `matrix-rust-sdk`: https://github.com/matrix-org/matrix-rust-sdk/pull/5737


src/error.rs
```diff
@@ -41,35 +41,57 @@ pub struct MegolmDecryptionError {
...
     }
+
+    /// Textual description of the withheld code, if any.
+    ///
+    /// Example: "The sender has disabled encrypting to unverified devices."
+    ///
+    /// `undefined` if we have not received a withheld code for the megolm
+    /// session.
+    #[wasm_bindgen(getter)]
+    pub fn maybe_withheld(&self) -> Option<String> {
```
uhoreg: I think this should have a `js_name` to make it camelCase in JavaScript.  Also, maybe call it `maybe_withheld_reason` to make it less easy to confuse with the `withheld_code` function

PR #273, opened 2025-11-03
Prepare sdk-crypto-wasm to use npm trusted publishing

richvdh:
Update to node 24 (and hence a sufficiently-recent npm), and request an
id-token from GHA.

https://github.com/element-hq/crypto-internal/issues/428


uhoreg (approved):
One issue, but looks plausible otherwise

PR #283, opened 2025-12-10
Update matrix-rust-sdk to 183116a4b1

andybalaam:
and adapt encrypt_room_event to preserve its interface despite the change introduced in https://github.com/matrix-org/matrix-rust-sdk/pull/5936 .

PR #284, opened 2025-12-11
Expose `forwarder` and `forwarderDevice` on `DecryptedRoomEvent`.

kaylendog:
Issue: https://github.com/element-hq/element-meta/issues/2877.

Depends on https://github.com/matrix-org/matrix-rust-sdk/pull/5945.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,14 +14,11 @@
     -   Added a new field `forwarder` to `InboundGroupSession` of type `ForwarderData`,
         which stores information about the forwarder of a session shared in a room key
         bundle under [MSC4268](https://github.com/matrix-org/matrix-spec-proposals/pull/4268).
-        ([#5980])(https://github.com/matrix-org/matrix-rust-sdk/pull/5980)
+        ([#5980](https://github.com/matrix-org/matrix-rust-sdk/pull/5980)
     -   The `OutboundGroupSession` and `OlmMachine` now return the `EncryptionInfo` used
         when encrypting raw events.
         ([#5936](https://github.com/matrix-org/matrix-rust-sdk/pull/5936))
-    -   Ensure that encrypted tests are run with a `StoreCipher`. This happened to reveal
-        tests which fail in an encrypted `EventCacheStore`, which required fixing queries
-        for all events in a room.
-        ([#5933](https://github.com/matrix-org/matrix-rust-sdk/pull/5933))
+        **Note:** The WASM bindings are not affected by this change.
 
 # matrix-sdk-crypto-wasm v16.0.0
 


src/responses.rs
```diff
@@ -244,6 +244,18 @@ pub struct EncryptionInfo {
...
+    /// event as part of an MSC4268 key bundle, if present. Only applicable for
+    /// room events.
+    #[wasm_bindgen(getter_with_clone)]
+    pub forwarder: Option<identifiers::UserId>,
+
+    /// The device ID of the user who sent us the keys with which we decrypted
+    /// this event as part of an MSC4268 key bundle, if present. Only applicable
+    /// for room events.
+    #[wasm_bindgen(getter_with_clone, js_name = "forwarderDevice")]
+    pub forwarder_device: Option<identifiers::DeviceId>,
```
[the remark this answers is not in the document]
richvdh: Right. Good question.

The thing is that I want to be able to get the forwarder details on the result of `decryptRoomEvent` without a separate call to `getRoomEventEncryptionInfo`. We can do that for the sender details today.

Anyway, it turns out I was somewhat confused by the getters on `DecryptedRoomEvent`: `sender`, `senderDevice`, `senderCurve25519Key`, etc. When I wrote that comment, I assumed that the `DecryptedRoomEvent` object actually had those fields: it turns out that it doesn't really, it just has getters that delegate to the nested `EncryptionInfo`.

You're right, `EncryptionInfo` is the right place: but we either need to add more getters to `DecryptedRoomEvent`, or just make its `EncryptionInfo` field public so that it is exposed to the JS side.

src/responses.rs
```diff
@@ -213,6 +213,25 @@ impl DecryptedRoomEvent {
     pub fn shield_state(&self, strict: bool) -> encryption::ShieldState {
         self.encryption_info.shield_state(strict)
     }
+
+    /// If the keys for this message were shared-on-invite as part of an
+    /// [MSC4268] key bundle, the ID of the user who sent us the bundle.
+    ///
+    /// [MSC4268]: https://github.com/matrix-org/matrix-spec-proposals/pull/4268
```
richvdh: bear in mind that this documentation will be interpreted as tsdoc, so this link style doesn't work: https://pr284--rust-sdk-crypto-wasm-doc-previews.netlify.app/classes/encryptioninfo#forwarder. Could you fix it before merging, please?

richvdh (commented):
ugh. Wait. can you mention the rust-sdk bump in the changelog too? (see previous changelog entries)

CHANGELOG.md
```diff
@@ -1,5 +1,32 @@
...
+        encrypted if they were shared as part of an [MSC4268](https://github.com/matrix-org/matrix-spec-proposals/pull/4268)
+        room key bundle.
+        ([#5945](https://github.com/matrix-org/matrix-rust-sdk/pull/5945)).
+    -   Added a new field `forwarder` to `InboundGroupSession` of type `ForwarderData`,
+        which stores information about the forwarder of a session shared in a room key
+        bundle under [MSC4268](https://github.com/matrix-org/matrix-spec-proposals/pull/4268).
+        ([#5980])(https://github.com/matrix-org/matrix-rust-sdk/pull/5980)
+    -   The `OutboundGroupSession` and `OlmMachine` now return the `EncryptionInfo` used
+        when encrypting raw events.
+        ([#5936](https://github.com/matrix-org/matrix-rust-sdk/pull/5936))
```
richvdh: This appears to refer to https://github.com/matrix-org/matrix-rust-sdk/pull/5936/changes#diff-9b100d6e2355b9e85edf44160e5883b7b9ec215262f0122475342bee7bb64c01R1099. AFAIK this info is not exposed outside the wasm crate (cf https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/283) so maybe don't mention? Or just explicitly say "(the interface to the wasm bindings are unaffected)" ?

CHANGELOG.md
```diff
@@ -1,5 +1,32 @@
...
+-   Expose information about the forwarder for events that were decrypted using a
+    key from an [MSC4268](https://github.com/matrix-org/matrix-spec-proposals/pull/4268)
+    key bundle.
+    ([#270](https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/284/changes))
+-   Update matrix-rust-sdk to `cd9f433`, which includes:
+
+    -   Fix `TimelineEvent::from_bundled_latest_event` sometimes removing the
+        `session_id` of UTDs. This broken event could later be saved to the event cache
+        and become an unresolvable UTD.
+        ([#5970](https://github.com/matrix-org/matrix-rust-sdk/pull/5970)).
```
richvdh: per discussion elsewhere, I don't think the wasm bindings use TimelineEvent, so I don't think this is relevant.

PR #287, opened 2026-01-15
Expose `hasDownloadedAllRoomKeys` property in crypto store.

kaylendog:
Exposes two new methods:

- `hasDownloadedAllRoomKeys` to access the property in the crypto store of the same name;
- `setHasDownloadedAllRoomKeys`, which adds the room with the given ID to the set in the crypto store.

Builds on functionality added in https://github.com/matrix-org/matrix-rust-sdk/pull/6017. Part of https://github.com/matrix-org/matrix-rust-sdk/issues/5111

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # UNRELEASED
 
+-   Expose two new methods, `has_downloaded_all_room_keys` and `set_has_downloaded_all_room_keys`
+    which allow clients to record whether they have downloaded all room keys from key backup for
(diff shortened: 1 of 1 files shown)


src/machine.rs
```diff
@@ -1902,6 +1902,43 @@ impl OlmMachine {
...
+        future_to_promise(async move {
+            me.store()
+                .save_changes(Changes {
+                    room_key_backups_fully_downloaded: HashSet::from_iter([room_id]),
+                    ..Default::default()
+                })
+                .await?;
+            Ok(JsValue::UNDEFINED)
+        })
+    }
```
richvdh: I'd stick these with the key backup stuff rather than the bundle-sharing stuff.

PR #291, opened 2026-02-04
Fix incorrect homepage and repository URLs

hughns:
https://github.com/matrix-org/matrix-rust-sdk-wasm is a 404, so I assume this just got left behind on an earlier rename?

I spotted it on npmjs.com:

image


richvdh (approved):
Thanks. It's had various names, but I don't think that was ever one of them.

PR #301, opened 2026-03-06
Switch package manager to `pnpm@10.3.0`

kaylendog:
This PR switches the project's package manager from `yarn` to `pnpm@10.3.0`, which brings it inline with a similar switch made by the `matrix-js-sdk` in https://github.com/matrix-org/matrix-js-sdk/pull/5184.

I have made three additional changes:

- Added an explicit dependency on `@types/node`, which I believe `yarn` erroneously included in `node_modules` as it is a dependency of `@types/jest`.
- Included `ES2022` in the `tsconfig` `libs` array, since `tests/helpers.ts` makes use of `Array.at`, which is an ES2022 feature.
- Set `noEmit` to `true` in `tsconfig`, since this is required by `allowImportingTsExtensions`.

I imagine we should split `tsconfig` into two, one for `node` and `browser`, but this is not a necessity.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,10 +26,10 @@ jobs:
 
     steps:
       - name: Checkout the repo
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Install Rust
-        uses: dtolnay/rust-toolchain@nightly
+        uses: dtolnay/rust-toolchain@0f1b44df7e9cbb178d781a242338dfa5e243ad7f # nightly
         with:
           components: rustfmt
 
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
 
     steps:
       - name: Checkout the repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v6
```
andybalaam: We should use commit IDs instead of version numbers to prevent malicious actors updating tags to inject malicious code. If this is too much hassle to include in the PR, feel free to defer until we have time,

andybalaam: (See https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions )

PR #308, opened 2026-03-11
docs: Update RELEASING.md to document correct pnpm usage

kaylendog:
#301 brought wasm-sdk inline with the js-sdk by switching to `pnpm`, but forgot to update `RELEASING.md`.


richvdh (approved):
lgtm otherwise

PR #310, opened 2026-03-13
Expose the vodozemac HPKE module

poljar:


src/vodozemac/hpke.rs
```diff
@@ -0,0 +1,325 @@
...
+#[wasm_bindgen(getter_with_clone)]
+pub struct BidirectionalCreationResult {
+    /// The fully established HPKE channel.
+    pub channel: EstablishedHpkeChannel,
+    /// The initial response.
+    #[wasm_bindgen(js_name = "initialResponse")]
+    pub initial_response: String,
+}
+
+/// An unestablished HPKE session.
```
richvdh: A general point on documentation: it's a bit hard to find this functionality via the generated documentation (https://pr310--rust-sdk-crypto-wasm-doc-previews.netlify.app/). It's maybe not a unique problem to hpke, but OlmMachine is at least mentioned in the example code.

Could we maybe add a section to the readme, that links to the generated docs for `HpkeRecipientChannel`?

I also wouldn't object to example code in the docs.

src/vodozemac/hpke.rs
```diff
@@ -0,0 +1,325 @@
...
+use matrix_sdk_crypto::vodozemac::hpke;
+use wasm_bindgen::prelude::*;
+
+use super::Curve25519PublicKey;
+
+fn used_up_error() -> JsError {
+    JsError::new("The HPKE channel was already established and used up.")
+}
+
+/// The result of an recipient HPKE channel establishment.
```
richvdh: A general comment for (almost) all the added structs: could we link back to the methods that return those structs, so that if you land on one in the documentation it is easier to navigate your way back up to the top level?

src/vodozemac/hpke.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// This method will use the
+    /// [`MATRIX_QR_CODE_LOGIN`](https://github.com/matrix-org/matrix-spec-proposals/pull/4108)
+    /// info for domain separation when creating the session.
+    #[wasm_bindgen(constructor)]
+    pub fn new() -> Self {
+        let inner = hpke::HpkeRecipientChannel::new();
+        let public_key = inner.public_key().into();
+
+        Self { inner: Some(inner), public_key }
+    }
```
richvdh: Note that we don't install a default tracing dispatcher, meaning any log lines from inside the rust will get swallowed.

Maybe that doesn't matter if the underlying Rust methods don't do any logging, but it might be confusing if we subsequently add logging.

The OlmMachine constructors allow you to pass in a Logger object, which is then used to create a dispatcher which is used for subsequent methods. Could we do the same here (and, presumably, clone the dispatcher into other structs such as the `UnidirectionalRecipientChannel` so they can do the same)?

src/vodozemac/hpke.rs
richvdh: Also also: could we mention that this is intended for use in an implementation of MSC4388?

richvdh (changes requested):
Looks generally good to me. A few comments.

src/vodozemac/hpke.rs
poljar: Like this https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/310/commits/30a84cb55dce5a932af65defa7f7ca8adaad21a8?

src/vodozemac/hpke.rs
poljar: I added a section to the readme in https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/310/commits/70cad40ef361f8433e9820bf557f79dc43ea3d84, the example was added in https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/310/commits/ff49ccd440b405585dd538725d5f9d0f0d2c65ed

poljar: Sure, add a couple of mentions in https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/310/commits/11305ef86269706c57ed2c4613b668466658f4d2.

src/vodozemac/hpke.rs
poljar: We can do this but,

> Maybe that doesn't matter if the underlying Rust methods don't do any logging, but it might be confusing if we subsequently add logging.

This hits the nail on the head. This re-exports a module from vodozemac. We had somewhat of a discussion if we'd want to vodozemac to ever emit logs.

We never let vodozemac emit logs, instead we expose enough info through result types and getters.

The HPKE support is a hybrid encryption scheme similar to what we did in the `PkEncryption` module of libolm. So it is a pretty simple protocol compared to the double ratchet.

Because of those two points, I don't expect this to ever emit any logs.

Now if you really want to avoid a breaking API change in the case my prediction is wrong, we can add this.

src/vodozemac/hpke.rs
```diff
@@ -16,6 +16,8 @@ fn used_up_error() -> JsError {
 }
 
 /// The result of an recipient HPKE channel establishment.
+///
+/// @see {@link HpkeRecipientChannel#establishChannel}
```
richvdh:
```suggestion
/// @see {@link HpkeRecipientChannel.establishChannel}
```

And likewise throughout.

richvdh: Personally, I don't find the formatting of the `@see` tag very useful; better just to write

```ts
/// See {@link HpkeRecipientChannel.establishChannel}.
```

YMMV

richvdh (approved):
could you s/#/./ in the links?

otherwise, lgtm, thanks!

src/vodozemac/hpke.rs
poljar: Changed the style in https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/310/commits/6c4231356d3af35a191c656057b0d6d65b6539e1.

PR #324, opened 2026-05-12
Release v19.0.0

poljar:


andybalaam (approved):

CHANGELOG.md
richvdh, line 11: Why the major version bump? If there are breaking changes, we should call them out in the changelog. If there aren't, we don't need a major version bump.

richvdh (changes requested):

CHANGELOG.md
poljar: We discussed this in chat.

The reason why I assumed a major version bump was required is because `PkEncryption::encrypt` method may throw an error now.

We don't consider this to be a breaking change, as such I'm going to close this and do a minor version bump instead.

PR #340, opened 2026-08-09
Fix `UserIdentity` methods to set up logging framework

richvdh:
The exposed methods on `OlmMachine` each set up the rust `tracing` system to send any emitted logs to the subscriber registered by the application. We'd forgotten to do the same for the `UserIdentity` methods, so things like `OtherUserIdentity::is_verified` weren't emitting any logs.


andybalaam (approved):
The guard mechanism looks like a hassle but I guess is required. Thanks!

PR #341, opened 2026-08-13
Support passing an async function to do X.509 signing

andybalaam:
Depends on https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/343

### DIFF
--- a/src/machine.rs
+++ b/src/machine.rs
@@ -6,6 +6,7 @@ use std::{
     iter,
     ops::Deref,
     pin::{pin, Pin},
+    sync::Arc,
     time::Duration,
 };
 
@@ -52,6 +53,7 @@ use crate::{
         SignatureVerification, StoredRoomKeyBundleData,
     },
     verification, vodozemac,
+    x509_signer::{wrap_x509_signer, X509SignFn, X509ValidityFn},
 };
 
 #[wasm_bindgen(typescript_custom_section)]
@@ -117,6 +119,17 @@ impl OlmMachine {
     ///   identities. Any user identity that has a valid signature according to
     ///   the supplied CAs will be considered verified, without any manual
     ///   verification taking place. A PEM-formatted string.
+    ///
+    /// * `raw_x509_signer` - Optional async function for signing some data with
+    ///   an X.509 certificate. Used to sign the user's identity so compatible
+    ///   clients will recognise this user as verified without manual
+    ///   verification taking place. If you supply this you must also supply
+    ///   raw_x509_validity.
+    ///
+    /// * `raw_x509_validity` - Optional function returning the validity period
+    ///   of the X.509 certificate used for signing, as the number of
+    ///   milliseconds since the Unix epoch. If you supply this you must also
+    ///   supply raw_x509_signer.
     #[wasm_bindgen(unchecked_return_type = "Promise<OlmMachine>")]
     pub fn initialize(
         user_id: &identifiers::UserId,
@@ -125,6 +138,13 @@ impl OlmMachine {
         #[wasm_bindgen(unchecked_optional_param_type = "string")] store_passphrase: Option<String>,
         #[wasm_bindgen(unchecked_optional_param_type = "JsLogger")] logger: Option<JsLogger>,
         #[wasm_bindgen(unchecked_optional_param_type = "String")] ca_certs_pem: Option<String>,
+        #[wasm_bindgen(
+            unchecked_optional_param_type = "(item: Uint8Array) => Promise<{ signature_bytes: Uint8Array, certificate_chain: string }>"
+        )]
+        raw_x509_signer: Option<X509SignFn>,
+        #[wasm_bindgen(unchecked_optional_param_type = "() => number")] raw_x509_validity: Option<
+            X509ValidityFn,
+        >,
     ) -> Promise {
         let dispatch = logger_to_dispatcher(logger);
         let _guard = dispatcher::set_default(&dispatch.clone());
@@ -133,7 +153,16 @@ impl OlmMachine {
         let device_id = device_id.inner.clone();
         future_to_promise(async {
             let store_handle = StoreHandle::open(store_name, store_passphrase).await?;
-            Self::init_helper(user_id, device_id, store_handle, dispatch, ca_certs_pem).await
+            Self::init_helper(
+                user_id,
+                device_id,
+                store_handle,
+                dispatch,
+                ca_certs_pem,
+                raw_x509_signer,
+                raw_x509_validity,
+            )
+            .await
         })
     }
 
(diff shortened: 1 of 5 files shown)


src/machine.rs
```diff
@@ -183,14 +258,154 @@ impl OlmMachine {
...
+            impl RawX509Signer for JsRawX509Signer {
+                fn sign(
+                    &self,
+                    message: Vec<u8>,
+                ) -> Pin<Box<dyn Future<Output = Result<RawX509Signature, SignatureError>>>>
+                {
+                    let signature_promise = self
+                        .signer_fn
+                        .call(&JsValue::NULL, (&Uint8Array::new_from_slice(&message),))
+                        .unwrap();
```
uhoreg: Can we get rid of this unwrap? (and the one below)

src/machine.rs
```diff
@@ -183,14 +258,154 @@ impl OlmMachine {
...
+                        Err(ValidityError)
+                    }
+                }
+            }
+
+            let signer = Arc::new(JsRawX509Signer {
+                signer_fn: raw_x509_signer,
+                validity_fn: raw_x509_validity,
+            });
+            builder.with_x509_signer(Some(signer))
```
richvdh: feels like this massive lump of code probably shouldn't be inline. Move it out to a function, with different definitions for the wasm and non-wasm case?

src/machine.rs
richvdh: can we avoid an unwrap() here?

src/machine.rs
```diff
@@ -183,14 +258,154 @@ impl OlmMachine {
...
+                {
+                    let signature_promise = self
+                        .signer_fn
+                        .call(&JsValue::NULL, (&Uint8Array::new_from_slice(&message),))
+                        .unwrap();
+
+                    let signature_future = JsFuture::from(signature_promise);
+
+                    Box::pin(async move {
+                        let signature: JsValue = signature_future.await.unwrap();
```
richvdh: ditto

src/machine.rs
andybalaam: Yep, sorry I'd forgotten about those - fix is 412ebdf

src/machine.rs
```diff
@@ -183,14 +258,154 @@ impl OlmMachine {
...
+
+                        let signature_bytes: Uint8Array =
+                            Reflect::get(&signature, &JsValue::from_str("signature_bytes"))
```
richvdh: did you consider using [`serde_wasm_bindgen::from_value`](https://docs.rs/serde-wasm-bindgen/latest/serde_wasm_bindgen/fn.from_value.html) with an appropriate rust-side type definition instead of these slightly awkward `Reflect` incantations?

src/machine.rs
```diff
@@ -75,6 +76,27 @@ pub struct OlmMachine {
...
+///
+/// Note: we assume (for now) that the signature scheme is RsaPssSha512, so
+/// no `signature_scheme` property is required.
+type X509SignFn = js_sys::Function<fn(Uint8Array) -> Promise<JsValue>>;
+
+/// The type of the user-provided JavaScript validity function that should
+/// return the certificate's validity period in seconds since the Unix epoch.
+///
+/// The returned value must be a number representing the number of seconds since
+/// the Unix epoch.
```
richvdh: all of our other time interfaces use milliseconds, not seconds. Suggest sticking to that for consistency.

richvdh (dismissed):
Needs a changelog entry, please

src/machine.rs
andybalaam: Done in 412ebdfa9fc4fdf4f5d0d614503cedb5eec445e8

src/machine.rs
andybalaam: Done in 6516b14

src/machine.rs
andybalaam: Yeah I did, but rejected it because it seemed like overkill, but thanks to your suggestion I had another look and it seems better. Fixed in f2f8cb2

src/machine.rs
andybalaam: Sorry, in retrospect this feels like I asked for review before it was ready. Split up and improved in 2602399 and 290a0ea

src/machine.rs
```diff
@@ -183,18 +238,35 @@ impl OlmMachine {
...
+            (Some(raw_x509_signer), Some(raw_x509_validity)) => {
+                // Wrap the provided X.509 signing function as a RawX509Signer to be used by the
+                // client to sign our identity.
+                //
+                // Note: if we are not on a WASM target, this function will panic, since we only
+                // support providing a signing function on WASM.
+                wrap_x509_signer(raw_x509_signer, raw_x509_validity, builder)
+            }
+            (Some(_), None) | (None, Some(_)) => {
+                panic!("You must supply both a raw_x509_signer and raw_x509_validity, or neither.")
```
uhoreg: Should this return an `Err` instead of panicking?  Since this function already returns a `Result`.

src/x509_signer.rs
```diff
@@ -0,0 +1,195 @@
...
+use serde::Deserialize;
+use wasm_bindgen::JsValue;
+use wasm_bindgen_futures::JsFuture;
+
+pub fn wrap_x509_signer(
+    raw_x509_signer: X509SignFn,
+    raw_x509_validity: X509ValidityFn,
+    builder: OlmMachineBuilder,
+) -> OlmMachineBuilder {
+    if !cfg!(target_family = "wasm") {
```
uhoreg: `!cfg!` looks odd.

```suggestion
    if cfg!(not(target_family = "wasm")) {
```

tests/machine.test.ts
```diff
@@ -253,6 +253,49 @@ eXmIj7ZEOIsufPdiYuKDp/aUdgUHmuCGyegfoCJze36SdFX5q6z8Aq5nKPtz+FM=
...
+                async (item: Uint8Array) => {
+                    return {
+                        signature_bytes: item,
+                        certificate_chain: "MY CHAIN",
+                        signature_scheme: "RsaPssSha512",
+                    };
+                },
+                () => 1000000,
+            ),
+        ).toBeInstanceOf(OlmMachine);
```
uhoreg, line 274: It would be nice to test that it will also sign our identity.  Though I'm not sure how hard it would be to get it to do that.

uhoreg (approved):
Looks pretty good.  A few small suggestions/changes, but I don't think it needs re-review after that.

src/machine.rs
andybalaam: Yeah, why not? 37ab940

tests/machine.test.ts
andybalaam: Yeah - I can't see a way to interrogate it for that info, but I'd love tests here that actually do something!

PR #348, opened 2026-09-15
Build with optimisation-level 3 and enable additional target features

hughns:
I'm not expecting that this PR is accepted "as is" and instead is an attempt to write up my findings so that the appropriate folks can pick up if of interest.

## TL;DR

Testing suggests this library could benefit from a >20% increase in performance, but at a cost of 35% increase in binary size.

## Background

Where https://github.com/matrix-org/matrix-rust-sdk/pull/5854 adjusted the the target-features for the Android ARM64 targets I wondered if there were any similar "low hanging fruit" for WASM targets.

I didn't find an existing benchmark to use so I asked Claude Opus for some help and together made https://github.com/matrix-org/matrix-js-sdk/tree/hughns/crypto-wasm-benchmark.

I wanted to run it in real browsers to understand any impact of the WASM engine and verify compatibility.

## Testing

I found that the [opt-level](https://docs.rs/opt-level/latest/opt_level/) and target features were available to play with.

The baseline [default features](https://doc.rust-lang.org/rustc/platform-support/wasm32-unknown-unknown.html#enabled-webassembly-features) are: `mutable-globals`, `nontrapping-fptoint`, `bulk-memory`, `sign-ext`, `reference-types`, `multivalue`.

The baseline optimisation level is `z` meaning: optimize for binary size, but also turn off loop vectorization.

Features available to play with that have good baseline support: [`simd128`](https://caniuse.com/wasm-simd), [`tail-call`](https://caniuse.com/wf-wasm-tail-call-optimization),  [`extended-const`](https://caniuse.com/wf-wasm-extended-constant-expressions)

The [`relaxed-simd`](https://caniuse.com/wf-wasm-simd-relaxed) flag does not have support on Safari yet.

Variants I tested with performance across Chromium/Firefox/WebKit are: (Negative = faster)

| test variant | [opt-level](https://docs.rs/opt-level/latest/opt_level/) | simd128 | tail-call | relaxed-simd | extended-const | chromium | firefox | webkit | **pooled** | Binary size (MiB) | FF compile time at startup |
|---|:-:|:-:|:-:|:-:|:-:|---:|---:|---:|---:|---:|---:|
| baseline | `z` | – | – | – | – | — | — | — | — | 7.47 | 91.4ms |
| simd128 | `z` | ✓ | – | – | – | +0.9% | −3.6% | −1.9% | −1.5% | 7.42 | 90.3ms |
| all | `z` | ✓ | ✓ | ✓ | ✓ | +1.6% | −3.0% | −1.7% | −1.1% | 7.42 | 88.0ms |
| os-all | `s` | ✓ | ✓ | ✓ | ✓ | −9.6% | −11.3% | −5.7% | **−8.9%** | 7.36 | 103.0ms |
| o2-all | `2` | ✓ | ✓ | ✓ | ✓ | −24.5% | −19.1% | −18.3% | **−20.7%** | 9.90 | 142.6ms |
| o3 | `3` | – | – | – | – | −25.2% | −19.2% | −18.5% | **−21.0%** | 10.14 | 146.1ms |
| o3-tailcall | `3` | – | ✓ | – | – | −24.0% | −19.7% | −18.0% | −20.6% | 10.17 | 153.4ms |
| o3-simd128 | `3` | ✓ | – | – | – | −25.2% | −20.8% | −18.4% | −21.5% | 10.02 | 162.4ms |
| o3-all | `3` | ✓ | ✓ | ✓ | ✓ | −24.8% | −22.9% | −19.5% | **−22.4%** | 10.02 | 176.6ms |

Steady state = `encrypt`+`decrypt` p50 geometric mean vs `baseline`, three browsers.
Noise floor (same build measured twice): −0.1%.

## Findings

- In my tests data the opt-level accounts for essentially all the gain (~−21%). Target features were must less significant.
- Variants `o3` (no optional features) and `o3-all` (all four) differ by 1.4 points; `o3-tailcall` is *slower* than plain `o3`. Isolating simd128 and tail-call via two independent paths each gave contradictory signs, so no individual feature has a defensible attribution.
- This could also be down to limitations of the test suite and device.
- `o2-all` and `o3-all` have identical feature flags yet differ by 1.7 points — the same magnitude as the entire feature effect.
- Opt-levels 2 and 3 are a rough tie (−20.7% vs −21.0%), but `-O2` is 0.24 MiB smaller and compiles 34ms faster on Firefox.
- Although Firefox startup time does rise, perhaps 100ms as a once off at startup time is acceptable.
- `os-all` is the size-conscious pick: −8.9% at 7.36 MiB, smaller than baseline.
- If file size is of concern then the "name" custom section of the binary (which I think contains debug strings for nice logs) accounts for 38% of the baseline binary size.


andybalaam:
Most WASM performance conversations talk a lot about size, because waiting for the WASM blob to download is a huge lump of dead time on the first time someone loads the application.

I think that has some relevance for us, but probably less than most web sites: Matrix clients are more of an "app" than a web site and generally expect users to stick around for a long time with a persistent tab, especially when using encrypted messaging.

Nevertheless, if we increase WASM size we should be mindful that will make the startup of any app slower. So the easiest to sell change would be one that makes the WASM smaller and makes execution faster. That seems like a no-brainer.

I also think experimenting with removing the "name" section is worth doing. I generally don't expect to be able to get much from debugging in the browser (although I definitely do want to know the line+file of a panic) but others may make more use of this info.

PR #351, opened 2026-09-22
Fix one of the async tests from failing

poljar:
Not sure why this passes on CI, but locally this fails for me as such:

```
FAIL  tests/asyncload.test.js
  ✕ can instantiate rust objects with async initialiser (1 ms)

  ● can instantiate rust objects with async initialiser

    ReferenceError: initUserId is not defined

      2 |
      3 | test("can instantiate rust objects with async initialiser", async () => {
    > 4 |     initUserId = () => new UserId("@foo:bar.org");
        |               ^
      5 |
      6 |     // stub out the synchronous WebAssembly loader with one that raises an error
      7 |     jest.spyOn(WebAssembly, "Module").mockImplementation(() => {

      at Object.<anonymous> (tests/asyncload.test.js:4:15)
```

I probably have a newer `pnpm` version?


andybalaam (approved):
Most odd.

